Issue #28, opened 2017-06-01
Search


nathaned:
Add coverage for Search

Only supporting _List All Courses_ since _Find recipients_ returns a list of mixed types.
https://canvas.instructure.com/doc/api/search.html#method.search.all_courses

Thetwam:
Resolved in #29 (despite some merging weirdness)

PR #29, opened 2017-06-02
Issue/28 search

nathaned:
Addresses #28 

Methods for both _Find recipients_ and _List all courses_ added, but functionality is limited as the data is difficult to work with.

Notes
---

- search_recipients() will return an empty list unless there is a search term
   - searching " " seems to include all (even results that do not contain spaces)
- both methods return response.json()

### DIFF
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -773,3 +773,38 @@ class Canvas(object):
             'folders/%s' % (folder_id)
         )
         return Folder(self.__requester, response.json())
+
+    def search_recipients(self, **kwargs):
+        """
+        Find valid recipients (users, courses and groups) that the current user
+        can send messages to.
+        Returns a list of mixed data types.
+
+        :calls: `GET /api/v1/search/recipients  \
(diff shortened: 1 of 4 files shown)


canvasapi/canvas.py
```diff
@@ -773,3 +773,41 @@ def get_folder(self, folder_id):
...
+        can send messages to.
+        Returns a list of mixed data types.
+
+        :calls: `GET /api/v1/search/recipients  \
+        <https://canvas.instructure.com/doc/api/search.html#method.search.recipients>`_
+
+        :rtype: `list`
+        """
+        if 'search' not in kwargs:
+            kwargs['search'] = ' '
```
Thetwam: What is the reason for this? Is the search keyword technically required?

Thetwam:
So... I'm not really sure what happened. When I hit the merge button, it initially appeared to work but GitHub isn't showing it as a successful merge. Confirmed that the changes are now in Develop.

Resolved in this commit: 46fb555607ebd6fc1216b4a1fdc08bc64c464bd0

PR #51, opened 2017-08-10
Expand SIS ID Support

### LINKED ISSUE #50: Query paramaters and sis ids
Hi, i am new to this api but I like the implementations of it a lot.

A lot of the canvas api accepts query parameters,
Lets take for example [listing users in a course](https://canvas.instructure.com/doc/api/live#!/courses.json/list_users_in_course_users_get_5) where almost every parameter type is query

I would like to be able to add a query parameters for example
/api/v1/courses/{course_id}/users?enrollment_type=student
Is this possible now or how would you suggest the implementation? 

Also, along with that it would also be handy to not use user_id but sis_user_id
and for course_id; sis_course_id. 
However, sis_ids are not a query string. But instead come with a "sis_course_id:" in front of the sis id number itself.
example:
```/api/v1/courses/sis_course_id:{{sis_course_id}}```

Is that possible?, suggest implementation?

Currently I seem to be able to make a custom query like this
This example fetches only students in a course and I have tested it to be working.
```
	canvas = Canvas(API_URL, API_KEY)

	# First, retrieve the Course object
	course = canvas.get_course(COURSE_ID)

	
	'''You can make your own custom request like this'''
	from canvasapi.paginated_list import PaginatedList
	from canvasapi.user import User

	st_users = PaginatedList(User,
					course._requester,
					'GET',
					'courses/{}/users?enrollment_type=student'.format(COURSE_ID)
					,{})
	for st in st_users:
		print(st.id,st.name)
```

### PR DESCRIPTION
**Pull request by Thetwam**
Added optional `use_sis_id` argument to methods for getting an account, course, group and section. Allows use of SIS IDs in appropriate id variables.

Resolves #50

### DIFF
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -380,7 +380,7 @@ class Canvas(object):
 
         :param group_id: The ID of the group to get.
         :type group_id: int or str
-        :param use_sis_id: Whether or not account_id is an sis ID.
+        :param use_sis_id: Whether or not group_id is an sis ID.
             Defaults to `False`.
         :type use_sis_id: bool
 
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -8,6 +8,7 @@ from canvasapi.folder import Folder
 from canvasapi.group import Group, GroupCategory
 from canvasapi.paginated_list import PaginatedList
 from canvasapi.requester import Requester
+from canvasapi.section import Section
 from canvasapi.user import User
 from canvasapi.util import combine_kwargs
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: canvasapi/canvas.py [hunk 1]
```diff
@@ -340,19 +370,30 @@ def create_group(self, **kwargs):
...
         """
         Return the data for a single group. If the caller does not
         have permission to view the group a 401 will be returned.
 
         :calls: `GET /api/v1/groups/:group_id \
         <https://canvas.instructure.com/doc/api/groups.html#method.groups.show>`_
 
+        :param group_id: The ID of the group to get.
+        :type group_id: int or str
+        :param use_sis_id: Whether or not account_id is an sis ID.
```
jessemcbride: This should be "whether or not `group_id` is an sis ID"

jessemcbride:
@Thetwam See comments on [line 383](https://github.com/ucfopen/canvasapi/pull/51/files#r132800159). Once that's addressed, this is clear for merge.

PR #67, opened 2017-09-03
Add Eclipse project metadata for use with PyDev

liblit:
These project settings files might be useful for anyone using Eclipse with [PyDev](http://www.pydev.org/).  Anyone else can just ignore them.

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>canvasapi</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.python.pydev.PyDevBuilder</name>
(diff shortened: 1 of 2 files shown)


jessemcbride:
I think we might want to refrain from including configuration for specific editors. I did some digging around to see how other projects handled it and I could only find things related to style guidelines (specifically [EditorConfig](http://editorconfig.org/), which I think we should look into!), not project setup.

@Thetwam Thoughts?

Thetwam:
I agree with @jessemcbride... Don't want to have too many niche configs. The EditorConfig thing *is* interesting, however. We may look into that in the future.

For the time being, I'm not opposed to adding `.project` and `.pydevproject` to the gitignore to make life easier for those using it.

jessemcbride:
Going to go ahead and close this then. Let's check out EditorConfig in a separate issue, though.

PR #70, opened 2017-09-05
Correct the git command used to create and check out a new branch

liblit:
`git branch` does not take a `-b` flag.  To create a new branch and check it out, one should use `git checkout -b`, not `git branch -b`.


jessemcbride:
Good catch. Thanks!

PR #87, opened 2017-09-29
Discussion Cleanup

### LINKED ISSUE #83: return type of Course.get_full_discussion_topic()
I'm struggling to understand the return of `get_full_discussion_topic()`. I would greatly appreciate if someone would give an example of how you might parse the output? The same question applies to `DiscussionTopic.list_topic_entries()`.

The endpoint `GET /api/v1/courses/:course_id/discussion_topics/:topic_id/view` natively returns a json string, which at least can be easily made into a dictionary. What can I do with the DiscussionTopic object type? Why was this chosen?

Thanks!

Thetwam:
Created a new class `DiscussionEntry`.
Updated several `DiscussionTopic` methods to return `DiscussionEntry` objects.
Moved several `DiscussionTopic` methods to the `DiscussionEntry` class.

Added new Exception `Forbidden` to handle HTTP 403 response
Fixed a few incorrect docstrings
Prepended underscores to `parent`  properties to prevent collisions and confusion with Canvas attributes of similar name.
String formatting tweaks

Resolves #83


jessemcbride:
Let's hold off on this until we have a discussion about `Unauthorized` vs `Forbidden` exceptions.

Thetwam:
Per our in-person discussion, adding `Forbidden` for 403 still seems appropriate. We may chose to (re)merge `InvalidAccessToken` and `Unauthorized` in the future, but that is outside of the scope of this issue in my opinion.

jackrsteiner:
Great. I am very excited to see these fixes! Thanks.

PR #90, opened 2017-10-02
Update Getting Started Page

### LINKED ISSUE #71: getting-started.rst Canvas object's order of arguments is inconsistent
I'm just starting out with both Python and canvasapi (and very excited about what you all have shared here!), and was using your getting-started guide to get up and going. I noticed this line at [the Usage section](https://github.com/ucfopen/canvasapi/blob/master/docs/getting-started.rst#usage):

`canvas = Canvas(API_URL, API_KEY)`

Then a few lines down the page under [Examples](https://github.com/ucfopen/canvasapi/blob/master/docs/getting-started.rst#examples), the order of the arguments is switched:

`canvas = Canvas(API_KEY, API_URL)` (repeated twice in the Examples section)

This could be confusing for beginners like me.

Thanks again for sharing your work! Looking forward to digging into the functionality some more.

Thetwam:
Updated "Getting Started" page in the documentation to match the updates to the README.

Also corrects the example of the `Course.update()` method.

Resolves #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Courses can be retrieved from the API:
 'Test Course'
 
 # Update the course's name
->>> course.update(name='New Course Name')
+>>> course.update(course={name: 'New Course Name'})
 ```
 
 See our documentation on [keyword arguments](#keyword-arguments) for more information about how `course.update()` handles the `name` argument.
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Courses can be retrieved from the API:
 'Test Course'
 
 # Update the course's name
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -72,7 +72,7 @@ Courses can be retrieved from the API:
 'Test Course'
 
 # Update the course's name
->>> course.update(name='New Course Name')
+>>> course.update(course={name: 'New Course Name'})
```
jessemcbride: I wish Python let us name keys without quotes.

`NameError: name 'name' is not defined`.

Issue #94, opened 2017-10-04
Handle other iterable sequences as "param[]=..." parameter values


liblit:
Issue #55 pointed out that `canvasapi` 0.6.0 made it awkward to pass a `list` as a parameter value. Pull request #63 fixed this, with friendly support for both `list` and `tuple` from `canvasapi` 0.7.0 forward. Yay! However, `list` and `tuple` are just two of the many Pythonic ways to collect multiple values. It would be nice if `canvasapi` provided the same parameter-marshaling support for any parameter value that is actually an iterable collection of multiple values: `set`, generators, iterable objects returned by `zip` and `ichain`, etc. Note, however, that strings are a special case: `str` is iterable, but should be treated as a single value, not as a sequence of isolated characters.

An example of something that works in `canvasapi` 0.7.0:

``` python
course.list_multiple_submissions(student_ids='all', assignment_ids=[1, 2, 3])
```

Some examples of things that do *not* work in `canvasapi` 0.7.0:

``` python
# multiple values as a set
assignment_ids = set(...)
course.list_multiple_submissions(student_ids='all', assignment_ids=assignment_ids)

# multiple values as an iterator over dictionary values
mapping = {'a': 1, 'b': 2, 'c': 3}
course.list_multiple_submissions(student_ids='all', assignment_ids=mapping.values())

# multiple values as a generator expression
course.list_multiple_submissions(student_ids='all', assignment_ids=(assn.id for assn in ...))
```

<hr/>

Spoiler alert: I am already developing code to implement this feature. Expect a pull request soonish.

Issue #55, opened 2017-08-25
Awkward to set context_codes[] parameter in Calendar.list_calendar_events call


liblit:
The `Canvas.list_calendar_events` method feeds its parameters down to a `GET /api/v1/calendar_events` request. [That request](https://canvas.instructure.com/doc/api/calendar_events.html#method.calendar_events_api.index) accepts a `context_codes[]` parameter, but setting that parameter correctly is surprisingly awkward in version 0.6.0 of `canvasapi`. The problem is that the parameter name really must be `context_codes[]`, not `context_codes`, but creating a Python keyword argument with brackets in its name is subtle:

```python
context = 'course_123456'

canvas.list_calendar_events(context_codes=context) 
# canvasapi.exceptions.CanvasException: API encountered an error processing your request

canvas.list_calendar_events(context_codes=[context])
# canvasapi.exceptions.CanvasException: API encountered an error processing your request

canvas.list_calendar_events(context_codes[]=context)
# SyntaxError: invalid syntax

canvas.list_calendar_events('context_codes[]'=context)
# SyntaxError: invalid syntax

canvas.list_calendar_events(**{'context_codes[]': context})
# success!
```

The syntactically invalid ones I knew were wrong before I even tried them. But for the first two attempts shown above, I guess I expected `canvasapi` to add the brackets for me either because it knew that `context_codes[]` was special, or perhaps it would always append brackets when I passed a value as a list (`[context]`) rather than a single item.

Arguably this is my fault: give the wrong parameter name, and of course things will fail. But I would have hoped that `canvasapi` would make this easier. The use of `**{...}` is obscure enough that many `canvasapi` users may not know how to do it at all.

Thetwam:
Hey liblit, thanks for reporting this issue!

Yeah, this is definitely not intentional behavior. Good idea to use `**{...}` syntax, that's a nice workaround!

Despite the documentation listing the type as `string`, it looks like it's supposed to accept a list. Because of that, in my mind, `context_codes=[context]` makes the most sense.

Do you think implementing it such that passing in a list as an argument auto-adds the brackets would be sufficient?

liblit:
You are most welcome, @Thetwam. Thank you for providing `canvasapi` in the first place! It has been a big help to me.

> Despite the documentation listing the type as string, it looks like it's supposed to accept a list

[...]

PR #95, opened 2017-10-05
Issue #94: Generalize repeated "param[]=..." support to all iterables

liblit:
Following the fix of issue #55 by pull request #63, canvasapi parameter handling treated any `list` or `tuple` as a sequence of multiple values.  When a caller provide multiple values in this way, they are be turned into repeated `param[]=...` parameters in an eventual Canvas API call.

The current pull request generalizes that support to *any* iterable collection of values, not just list and tuple.  One can now pass multiple values using `list` or `tuple` as before, but also `set`, generators, iterable objects returned by `zip` and `ichain`, etc. Strings and byte sequences are special cases.  Although these are iterable, each is treated a single value rather than as a sequence of isolated characters or bytes.

If accepted, this pull request will resolve issue #94.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -65,6 +65,35 @@ class TestUtil(unittest.TestCase):
             < result.index(('foo[]', 'bar4'))
         )
 
+    def test_combine_kwargs_single_generator_empty(self, m):
+        result = combine_kwargs(var=(value for value in ()))
+        self.assertIsInstance(result, list)
+        self.assertEqual(len(result), 0)
+
+    def test_combine_kwargs_single_generator_single_item(self, m):
+        result = combine_kwargs(var=(value for value in ('test',)))
+        self.assertIsInstance(result, list)
+        self.assertEqual(len(result), 1)
+        self.assertIn(('var[]', 'test'), result)
+
+    def test_combine_kwargs_single_generator_multiple_items(self, m):
+        result = combine_kwargs(foo=(value for value in ('bar1', 'bar2', 'bar3', 'bar4')))
+        self.assertIsInstance(result, list)
+        self.assertEqual(len(result), 4)
+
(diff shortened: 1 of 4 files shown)


liblit:
Excellent suggestion, @Thetwam. I will comply when time permits.

tests/test_util.py
```diff
@@ -6,17 +6,89 @@
...
-from canvasapi.util import combine_kwargs, obj_or_id
+from canvasapi.util import combine_kwargs, is_multivalued, obj_or_id
+from itertools import chain
+from six import integer_types, iterkeys, itervalues, iteritems
 from tests import settings
 from tests.util import register_uris
 
+try:
+    # in Python 2.x, iterator-returning zip function is "itertools.izip"
+    from itertools import izip as zip
```
Thetwam: `six` provides this functionality as `six.moves.zip`. https://pythonhosted.org/six/
in py2 it's `itertools.izip()` and in py3 it's just `zip`

Thetwam:
> many (too many?) unit tests
never enough 😄

tests/test_util.py
```diff
@@ -6,17 +6,89 @@
...
+    # is_multivalued()
+    def test_is_multivalued_bool(self, m):
+        self.assertFalse(is_multivalued(False))
+
+    def test_is_multivalued_integer_types(self, m):
+        for type in integer_types:
+            self.assertFalse(is_multivalued(type(1)))
+
+    def test_is_multivalued_str(self, m):
+        self.assertFalse(is_multivalued('string'))
```
Thetwam: Since we import `unicode_literals`, this will always end up being a unicode string in both py2 and py3.

Not sure if we should remove it or leave it in. Certainly doesn't hurt to leave it.

PR #111, opened 2017-11-14
Issue/89 notification preferences update

a-goetz:
Added handling of additional notification preference endpoints.

Remain to be seen if issue number 89 is resolved, since there are still uncovered endpoints.


canvasapi/communication_channel.py
```diff
@@ -66,11 +70,98 @@ def get_preference(self, notification):
...
+        kwargs['notification_preferences[frequency]'] = frequency
+        response = self._requester.request(
+            'PUT',
+            'users/self/communication_channels/{}/notification_preferences/{}'.format(
+                self.id,
+                notification
+            ),
+            _kwargs=combine_kwargs(**kwargs)
+        )
+        data = response.json()['notification_preferences'][0]
```
jessemcbride: I would be wary of list access like this. Can it be guaranteed that we won't get an `IndexError` trying to access the first element?

canvasapi/communication_channel.py
```diff
@@ -66,11 +70,98 @@ def get_preference(self, notification):
...
         response = self._requester.request(
             'GET',
-            'users/%s/communication_channels/%s/notification_preferences/%s' % (
+            'users/{}/communication_channels/{}/notification_preferences/{}'.format(
                 self.user_id,
                 self.id,
                 notification
             )
         )
         data = response.json()['notification_preferences'][0]
```
jessemcbride: See my comment below. We should have a discussion about the potential side effects of this line.

canvasapi/communication_channel.py
a-goetz: It should work in this case. Canvas is doing a new thing with returning lists, we probably need to create something to handle it.

tests/fixtures/notification_preference.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
Thetwam: What's up with this file? Should it be here?

PR #116, opened 2017-11-28
Implemented Grading Standards

JonGuilbe:
Created 3 methods for Grading Standards in account.py and course.py.
Added tests for the added methods.
Added Visual Studio Code settings file to .gitignore.

### DIFF
--- a/canvasapi/account.py
+++ b/canvasapi/account.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 from six import python_2_unicode_compatible
 
 from canvasapi.canvas_object import CanvasObject
+from canvasapi.grading_standard import GradingStandard
 from canvasapi.exceptions import RequiredFieldMissing
 from canvasapi.paginated_list import PaginatedList
 from canvasapi.util import combine_kwargs, obj_or_id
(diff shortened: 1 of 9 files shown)


canvasapi/account.py
```diff
@@ -1039,6 +1040,74 @@ def get_all_outcome_links_in_context(self):
...
+        response = self._requester.request(
+            'POST',
+            'accounts/%s/grading_standards' % (self.id),
+            title=title,
+            _kwargs=combine_kwargs(**kwargs)
+        )
+
+        return GradingStandard(self._requester, response.json())
+
+    def get_grading_standards(self):
```
Thetwam: Add `**kwargs` to this and other methods. Although the Canvas documentation doesn't list any arguments, anything returning a paginated list can benefit from the user passing args such as `per_page`. To be safe, add it to all new methods you've created here.

Be sure to pass it along w/ combine_kwargs!

canvasapi/grading_standard.py
```diff
@@ -0,0 +1,12 @@
...
+from six import python_2_unicode_compatible
+
+from canvasapi.canvas_object import CanvasObject
+
+
+@python_2_unicode_compatible
+class GradingStandard(CanvasObject):
+
+    def __str__(self):
+        return "{}".format(self.grading_scheme)
```
Thetwam: Instead of grading_scheme, use the title with the id in parentheses like this:

>Account Standard (1)

tests/fixtures/account.json
```diff
@@ -1004,5 +1004,65 @@
...
+			"context_type": "Account",
+			"grading_scheme": [
+			  {"name": "A", "value": 0.9},
+			  {"name": "B", "value": 0.8},
+			  {"name": "C", "value": 0.7}
+			]
+		  },
+		  {
+			"title": "Grading Standard 2",
+			"id": 1,
```
Thetwam: Update IDs to prevent duplicates. Check other fixtures too.

tests/fixtures/course.json
```diff
@@ -1391,5 +1391,65 @@
...
+			"grading_scheme": [
+			  {"name": "A", "value": 0.9},
+			  {"name": "B", "value": 0.8},
+			  {"name": "C", "value": 0.7}
+			]
+		  },
+		  {
+			"title": "Grading Standard 2",
+			"id": 2,
+			"context_id": 2,
```
Thetwam: This `context_id` combines with `context_type` to reference the parent object, in this case Course #1. Update `context_id` to reflect this.

tests/test_account.py
```diff
@@ -623,3 +624,71 @@ def test_get_outcome_links_in_context(self, m):
...
+        response = self.account.add_grading_standards(title, grading_scheme)
+
+        self.assertIsInstance(response, GradingStandard)
+        self.assertTrue(hasattr(response, 'title'))
+        self.assertEqual(title, response.title)
+        self.assertTrue(hasattr(response, "grading_scheme"))
+        self.assertEqual(response.grading_scheme[0].get('name'), "A")
+        self.assertEqual(response.grading_scheme[0].get('value'), 0.9)
+
+    # add_grading_standards()
```
Thetwam: No need to duplicate comments here. All `add_grading_standards` tests should be under a single comment.

tests/fixtures/course.json
```diff
@@ -1391,5 +1391,65 @@
...
+		  }
+		],
+		"status_code": 200
+	},
+	"get_single_grading_standard": {
+		"method": "GET",
+		"endpoint": "courses/1/grading_standards/1",
+		"data": {
+			"title": "Grading Standard 1",
+			"id": 4,
```
Thetwam: In this case, the request is to `courses/1/grading_standards/1` which means the grading standard ID should be 1, not 4.

tests/test_account.py
```diff
@@ -623,3 +624,68 @@ def test_get_outcome_links_in_context(self, m):
...
+
+    # get_single_grading_standards()
+    def test_get_single_grading_standard(self, m):
+        register_uris({'account': ['get_single_grading_standard']}, m)
+
+        response = self.account.get_single_grading_standard(1)
+
+        self.assertIsInstance(response, GradingStandard)
+        self.assertTrue(hasattr(response, 'id'))
+        self.assertEqual(4, response.id)
```
Thetwam: This ties into the change made in fixtures... On line 682 you made a call to get a grading standard with an ID of 1 (not 4). That should be reflected here.

tests/fixtures/course.json
```diff
@@ -1391,5 +1391,65 @@
...
+		  },
+		"status_code": 200
+	},
+	"get_grading_standards": {
+		"method": "GET",
+		"endpoint": "courses/1/grading_standards",
+		"data": [
+			{
+			"title": "Grading Standard 1",
+			"id": 2,
```
Thetwam: It's perfectly valid for this to have an ID of 2, however changing it from 1 to 2 was unnecessary. My original comment on ID numbers was to change the ID of "Grading Standard 2" so it didn't have the same ID as "Grading Standard 1" (they were previously both 1)

PR #117, opened 2017-11-28
#115 Deprecate api version in url

### LINKED ISSUE #115: Deprecate /api/v1 from Canvas constructor
Currently we require `/api/v1/` to be appended to the end of an institution's Canvas URL when initializing the `Canvas` object.

We don't check if the URL is malformed in anyway, and as a result, sometimes unexpected behavior can present itself. We recently ran into this issue with one of our own applications where the base URL was set to `instructure.edu//api/v1/`.

We should implement a fix for this in two ways:

* Remove the `/api/version` requirement in the URL as we don't currently support targeting more than one API version
* Force URLs to be formatted properly (i.e. `*.someurl.com` with no trailing slash, `https://` or `http://` allowed) and manually append the `/api/v1/` string ourselves afterward

We should also warn the user that `/api/v1/` has been deprecated and is no longer required if they include it in their base URL.

jessemcbride:
Resolves #115

### DIFF
--- a/canvasapi/util.py
+++ b/canvasapi/util.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from collections import Iterable
 from six import binary_type, string_types, text_type
-from six.moves.urllib_parse import urlparse
 
 
 def is_multivalued(value):
@@ -127,12 +126,15 @@ def obj_or_id(parameter, param_name, object_types):
 
 def get_institution_url(base_url):
     """
-    Given an API root URL, grab the protocol (http/https) and the host
-    and then return that, discarding the rest of the string.
+    Trim '/api/v1' from a given root URL.
 
     :param base_url: The base URL of the API.
     :type base_url: str
     :rtype: str
     """
-    parsed = urlparse(base_url)
-    return parsed.scheme + '://' + parsed.netloc
+    index = base_url.find('/api/v1')
+
+    if index != -1:
+        return base_url[0:index]
+
+    return base_url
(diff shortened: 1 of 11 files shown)


jessemcbride:
Looks like I should have run this on Python 3 before opening the PR. `urlparse` was moved into `urllib` in Python 3. Need to use six to make them compatible for both versions before this can be merged.

canvasapi/util.py
```diff
@@ -122,3 +123,16 @@ def obj_or_id(parameter, param_name, object_types):
...
+    """
+    Given an API root URL, grab the protocol (http/https) and the host
+    and then return that, discarding the rest of the string.
+
+    :param base_url: The base URL of the API.
+    :type base_url: str
+    :rtype: str
+    """
+    parsed = urlparse(base_url)
+    return parsed.scheme + '://' + parsed.netloc
```
Thetwam: This feels heavy-handed to me. It's a pretty huge assumption that every instance of Canvas ever will be exactly at the root of a site.

It's entirely possible that a University's Canvas instance could live at `https://fakeuniversity.edu/canvas`, or something similar. This method would (incorrectly) cut off the `/canvas` bit. Even if we assume all institutions do follow the same url pattern, there's no guarantee that every self-hosted development instance will.

I think we should instead specifically trim `/api/v1/`, `/api/v1`, etc. from the url instead of dropping the path.

Thetwam: There's also a weird side effect where if you pass an invalid url to this function:

```
>>> get_institution_url('')
u'://'
>>> get_institution_url('test')
u'://'
>>> get_institution_url('google.com')
u'://'
```

Perhaps raise an exception if it's not valid?

jessemcbride: Agreed on using `netloc`. I've updated this to remove urlparse, now it'll just trim the end of a string containing /api/v1.

Issue #81, opened 2017-09-21
Quiz Questions and Question Groups


Thetwam:
Add coverage for Quiz Questions and Quiz Question Groups

https://canvas.instructure.com/doc/api/quiz_questions.html
https://canvas.instructure.com/doc/api/quiz_question_groups.html

liblit:
I believe that this issue was closed prematurely. The [original request](#issue-259588741) mentioned both quiz questions and quiz question groups. @jessemcbride’s #118 may have added support for quiz question groups, but I see no support for the [Quiz Questions API](https://canvas.instructure.com/doc/api/quiz_questions.html).

Thetwam:
Whoops! Thanks @liblit. Reopening this one.

Still need to do Quiz Questions:
https://canvas.instructure.com/doc/api/quiz_questions.html

PR #118, opened 2017-11-30
Issue/81 quiz question groups

JonGuilbe:
Created 5 Quiz Question Group methods in course.py.
Created quiz_group.py.
Created tests and added to fixtures for all created methods.


canvasapi/course.py
```diff
@@ -1659,6 +1660,167 @@ def get_outcome_result_rollups(self, **kwargs):
...
+        :type: `int`
+        :param id: The ID of the question group.
+        :type: `int`
+
+        :returns: `QuizGroup` object
+        :rtype: :class:`canvasapi.quiz_group.QuizGroup`
+        """
+        response = self._requester.request(
+            'GET',
+            'courses/{}/quizzes/{}/groups/{}'.format(self.id, quizid, id)
```
Thetwam: Looking at the URL this hits, it should be in the `Quiz` object, not `Course`.

canvasapi/course.py
```diff
@@ -1659,6 +1660,167 @@ def get_outcome_result_rollups(self, **kwargs):
...
+        if ("name" not in quiz_groups[0] and "pick_count" not in quiz_groups[0]
+            and "question_points" not in quiz_groups[0]
+                and "assessment_question_bank_id" not in quiz_groups[0]):
+            raise RequiredFieldMissing("quiz_groups must contain at least 1 parameter.")
+
+        kwargs["quiz_groups"] = quiz_groups
+
+        response = self._requester.request(
+            'POST',
+            'courses/{}/quizzes/{}/groups'.format(self.id, quizid),
```
Thetwam: Method should be in `Quiz` object

canvasapi/course.py
```diff
@@ -1659,6 +1660,167 @@ def get_outcome_result_rollups(self, **kwargs):
...
+            _kwargs=combine_kwargs(**kwargs)
+        )
+        return QuizGroup(self._requester, response.json().get('quiz_groups')[0])
+
+    def update_question_group(self, quizid, id, quiz_groups, **kwargs):
```
Thetwam: This method should probably be on the `QuizGroup` object. Probably just called `update`

canvasapi/course.py
```diff
@@ -1659,6 +1660,167 @@ def get_outcome_result_rollups(self, **kwargs):
...
+        kwargs["quiz_groups"] = quiz_groups
+
+        response = self._requester.request(
+            'PUT',
+            'courses/{}/quizzes/{}/groups/{}'.format(self.id, quizid, id),
+            _kwargs=combine_kwargs(**kwargs)
+        )
+        return QuizGroup(self._requester, response.json().get('quiz_groups')[0])
+
+    def delete_question_group(self, quizid, id):
```
Thetwam: Should probably be on `QuizGroup` object as `delete`

canvasapi/course.py
```diff
@@ -1659,6 +1660,167 @@ def get_outcome_result_rollups(self, **kwargs):
...
+        :returns: True if the result was successful (Status code of 204)
+        :rtype: `bool`
+        """
+        response = self._requester.request(
+            'DELETE',
+            'courses/{}/quizzes/{}/groups/{}'.format(self.id, quizid, id)
+        )
+        return (response.status_code == 204)
+
+    def reorder_question_group(self, quizid, id, order, **kwargs):
```
Thetwam: Should be on `QuizGroup` object

canvasapi/quiz.py
```diff
@@ -50,3 +52,69 @@ def delete(self, **kwargs):
...
+        """
+        Create a new question group for the given quiz id
+
+        :calls: `POST /api/v1/courses/:course_id/quizzes/:quiz_id/groups/:id \
+        <https://canvas.instructure.com/doc/api/quiz_question_groups.html#method.quizzes/quiz_groups.create>`_
+
+        :param quiz_groups: The name, pick count, question points,
+            and/or assessment question bank id.
+            All of these parameters are optional, but at least one must exist
+            (even if empty) to recieve a response.
```
jessemcbride: `receive` not recieve

canvasapi/quiz.py
```diff
@@ -50,3 +52,69 @@ def delete(self, **kwargs):
...
+            The request expects a list, but will only create 1 question group per request.
+        :type quiz_groups: list[dict]
+
+        :returns: `QuizGroup` object
+        :rtype: :class:`canvasapi.quiz_group.QuizGroup`
+        """
+
+        if not isinstance(quiz_groups, list) or len(quiz_groups) <= 0:
```
jessemcbride: A list (`quiz_groups`) cannot contain less than one element. When a list is evaluated as a boolean (e.g. a conditional like this), an empty list evaluates to `False`.

```
>>> test = []
>>> bool(test)
False
```

So we can replace `len(quiz_groups) <= 0` with `quiz_groups`.

canvasapi/quiz_group.py
```diff
@@ -0,0 +1,111 @@
...
+            'PUT',
+            'courses/{}/quizzes/{}/groups/{}'.format(self.course_id, self.quiz_id, id),
+            _kwargs=combine_kwargs(**kwargs)
+        )
+
+        successful = 'name' in response.json().get('quiz_groups')[0]
+        if successful:
+            super(QuizGroup, self).set_attributes(response.json().get('quiz_groups')[0])
+
+        return 'name' in response.json().get('quiz_groups')[0]
```
jessemcbride: This expression is already cached in line 52.

`return successful` instead of making Python look up the value again.

canvasapi/quiz_group.py
```diff
@@ -0,0 +1,111 @@
...
+        :type id: int
+        :param order: A list of dictionaries containing the key 'id' of
+            the question to be placed at order's index.
+        :type id: list[dict]
+
+        :returns: True if the result was successful (Status code of 204)
+        :rtype: bool
+        """
+
+        if not isinstance(order, list) or len(order) <= 0:
```
jessemcbride: See my previous comments about `len(order) <= 0`.

canvasapi/quiz_group.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ValueError("Param `order` must be a non-empty list.")
+
+        for orderDict in order:
```
jessemcbride: We're writing Python! Use snake_case not camelCase. 

I'm also not sold on `order_dict` as a variable name. Is there something more descriptive that doesn't include the variable type?

canvasapi/quiz.py
Thetwam: Directly evaluating as a boolean is good, though I'm going to replace `len(quiz_groups) <= 0` with `not quiz_groups` instead of `quiz_groups`, as the desired result is for the expression to evaluate to `True` when the list is empty.

canvasapi/quiz_group.py
Thetwam: Good catch. Looks like I didn't fully convert 😊

jessemcbride:
Lgtm

Thetwam:
Resolves #81

Issue #119, opened 2017-12-21
File Download


DanBrink91:
Add a method for downloading files to the File object

jessemcbride:
Per our conversations about this in the office, I'm thinking we should have two methods: one to provide a download that writes directly to the machine and one that allows you to download it and keep it in memory.

Thetwam:
Do those need to be separate methods? The code will be largely the same for both, just depends where it puts the file. I feel like an optional `path` argument would be sufficient. If provided, write the file to the path (and maybe return the handler?). Otherwise, just return the file directly.

PR #120, opened 2017-12-21
Issue/119 File Downloads

DanBrink91:
Putting this up so we can start a discussion on how to mock a file download and if we need anything else from this method.

Solves issue #119


DanBrink91:
Updated with tests

canvasapi/file.py
```diff
@@ -25,3 +25,29 @@ def delete(self):
...
+        """
+        response = self._requester.request(
+            'GET',
+            _url=self.url
+        )
+        return response.text
+
+    def download(self, location):
+        """
+        Download the file to specificed location
```
jessemcbride: specified* 

End the docstring with a period too pls

canvasapi/file.py
```diff
@@ -25,3 +25,29 @@ def delete(self):
...
+            'GET',
+            _url=self.url
+        )
+        return response.text
+
+    def download(self, location):
+        """
+        Download the file to specified location.
+
+        :param location: The path to download to.
```
jessemcbride: Include the type's parameter after the `:param location:`.

```
:type location: str
```

tests/test_file.py
```diff
@@ -7,6 +7,8 @@
 from canvasapi.file import File
 from tests import settings
 from tests.util import register_uris
+from tests.util import cleanup_file
```
jessemcbride: Flip import order to make it alphabetical.

tests/test_file.py
```diff
@@ -35,3 +37,19 @@ def test_delete_file(self, m):
...
+
+    # download()
+    def test_download_file(self, m):
+        register_uris({'file': ['file_download']}, m)
+
+        self.file.download('canvasapi_file_download_test.txt')
+        self.assertTrue(isfile('canvasapi_file_download_test.txt'))
+        with open('canvasapi_file_download_test.txt') as downloaded_file:
+            self.assertEqual(downloaded_file.read(), '"file contents are here"')
+        cleanup_file('canvasapi_file_download_test.txt')
```
Thetwam: It's technically possible (though unlikely) for this to not run if part of the tests fails. That means the file wouldn't be cleaned up and would be left on the test runner's system. To get around this in other places, we've wrapped stuff in a `try`/`finally`.

PR #137, opened 2018-01-16
Change list_users to GET, was wrongly marked POST

sigurdurb:


jessemcbride:
Thanks for catching this, @sigurdurb. 

I just noticed that there's more wrong with that docstring than just the HTTP method. The parameter `invitees` and its type are also mentioned despite not being a valid argument for that call.

I'll go ahead and make that change and then merge this in.

PR #159, opened 2018-02-27
Fix Submission Methods

### LINKED ISSUE #57: Submission methods on wrong class
Was working with assignment submissions recently and discovered that many of the methods relating to submissions are in the wrong class. For example, `list_submissions`appears in the `Course` and `Section` classes, but should probably just be in the `Assignment` class.

Need to move them to the appropriate class.

Thetwam:
Moved several methods from Course and Section to Submission and Assignment. Deprecated old methods. Resolves #57

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
   - 2.7
-  - 3.3
   - 3.4
   - 3.5
   - 3.6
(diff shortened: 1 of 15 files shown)


Thetwam:
Noticed some weirdness in the tests on py3.3 ... investigating...

Issue #163, opened 2018-03-03
Wrap "Copy a file" method


liblit:
The Canvas REST API offers a ["Copy a file" method](https://canvas.instructure.com/doc/api/files.html#method.folders.create_file) that `canvasapi` 0.9.0 does not yet wrap. I recently had cause to use this, so a proper `canvasapi` version of this would be appreciated. Presumably this makes the most sense as a method of the `Folder` class.

PR #165, opened 2018-03-07
#163 Copy File Method in Folder

qwertynerd97:


Thetwam:
This PR contains commits that are already in #164, so will wait until that PR is merged before reviewing.

Issue #156, opened 2018-02-21
Standardize Method Names


Thetwam:
We have several endpoints that are prefixed with `list_` while others are prefixed with `get_`. For consistency's sake, we need to standardize which form to use. After internal discussion, we have decided that `get_` seems the most appropriate.

Change all method names that start with `list_` to start with `get_`.

Because some folks may already be using the `list_` methods, we can't just rename the methods - we'll need deprecate them and direct users to switch to the new `get_` methods. Most likely move the code to the `get_` version and have the `list_` version warn with a `DeprecationWarning` and call the `get_` version.

PR #166, opened 2018-03-09
#156 Standardize Method names

qwertynerd97:


tests/test_outcome.py
```diff
@@ -203,17 +204,47 @@ def test_list_linked_outcomes(self, m):
...
+            self.assertIsInstance(result[0], OutcomeLink)
+            self.assertEqual(result[0].outcome_group['id'], 2)
+            self.assertEqual(result[0].outcome_group['title'], "Global Test Outcome Group")
+
+            result = self.course_outcome_group.list_linked_outcomes()
+            self.assertIsInstance(result[0], OutcomeLink)
+            self.assertEqual(result[0].outcome_group['id'], 2)
+            self.assertEqual(result[0].outcome_group['title'], "Course Test Outcome Group")
+
+            self.assertEqual(len(warning_list), 1)
```
Thetwam: Python 2 and Python 3 handle warnings differently. Python 3 doesn't consider duplicates of the same warning when catching and counting.

In this test's case, the DeprecationWarning is raised 3 times, but caught only once in Python 3.

In Python 2, a warning is counted each time it is caught. This means this line will raise `AssertionError: 3 != 1`, as 3 separate warnings are raised.

Issue #138, opened 2018-01-17
Correct erroneous docstrings


jessemcbride:
We need to crawl through our docstrings and verify that they're all correct for the endpoint they're calling. Though I would imagine the number of incorrect docstrings is small, we've noticed a few here and there and I'd feel better doing a thorough search of our codebase.

Bonus points if someone writes a script to do this automatically (please? 😊 ).

MarkLalor:
As far as a script is concerned, you mean something that checks, for example, that
```
:calls: `GET /api/v1/courses/:course_id/students/submissions \
```
matches the subsequent call in the implementation?
```
 'GET',
 'courses/{}/students/submissions'.format(self.id),
```

It would be nice if it also checked for dead links to the API documentation, i.e. in this case it would make sure `https://canvas.instructure.com/doc/api/submissions.html#method.submissions_api.for_students` still exists (both the page and the header)

jessemcbride:
Yes, that's what I was thinking. I don't think it would require much magic beyond [inspect](https://docs.python.org/2/library/inspect.html) and some regex. We've been looking for ways to automate the generation of wrapper methods using Canvas's documentation and I think this would be a decent first step.

Thetwam:
Resolved in #179. Still a few outstanding issues, but on the whole is good. Remaining issues are in #186

PR #179, opened 2018-05-08
#138 correct erroneous docstrings: corrections and script to find inconsistencies

rmanbaird:


Thetwam:
Addresses #138

Thetwam:
The changes you've made are good. The script is awesome, too! Looks like there are a few outstanding docstring issues, primarily related to Accounts. It appears the API has changed since our implementation of these methods, so they need be looked at. I'll spin those out into their own issue(s) to be resolved later.

Good work! 👍

PR #194, opened 2018-06-27
Coveralls to Codecov

Thetwam:
Changed default coverage runner from coveralls to codecov


Thetwam:
Codecov is pretty cool, but I'm having a hard time getting it to work on the ucfopen repo. (Despite it working fine on my local)

Has to do with the way they handle admin permissions - it only takes into account org-level permissions, not repo-level. Doesn't seem worth the hassle.

PR #196, opened 2018-06-30
Add uploading a file to a submission for an assignment

### LINKED ISSUE #195: Add coverage for uploading a file to a submission for an assignment.
Currently, submitting in the context of an assignment is not covered. This endpoint is documented [here](https://canvas.instructure.com/doc/api/submissions.html#method.submissions_api.create_file)

This is really useful to use as a student.

MarkLalor:
Resolves #195


MarkLalor:
Thanks for picking this up, I sort of forgot about this PR. I agree with all your suggestions, really like  the `obj_or_id` addition. I think an optional argument to upload automatically makes sense. Although it might be more clear if instead of an optional argument, it's another method like `upload_and_submit` or more even specific `upload_to_submission_and_submit`.

Thetwam:
To add another method that does the same thing - but with one additional feature - seems redundant. We'd essentially be copy-pasting the code from `submit()` and just adding the file upload to the new method.

I've taken a stab at the optional argument implementation. Thoughts?

canvasapi/assignment.py
```diff
@@ -141,17 +141,15 @@ def submit(self, submission, file=None, **kwargs):
...
 
             upload_response = self.upload_to_submission(file, **kwargs)
             if upload_response[0]:
                 kwargs['submission']['file_ids'] = [upload_response[1]['id']]
             else:
                 raise CanvasException(
-                    'File upload failed; Not submitting.\n Response: {}'.format(
-                        upload_response[1]
-                    )
+                    'File upload failed. Not submitting.'.format()
```
jessemcbride: Do we need the format call here?

Thetwam: uhh, whoops. Fixed.

Issue #216, opened 2018-10-22
Assignment Overrides


Thetwam:
We're missing coverage for Assignment Overrides. Add coverage for the following endpoints:

- [x] [List assignment overrides](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.index)
- [x] [Get a single assignment override](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.show)
- [x] [Redirect to the assignment override for a group](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.group_alias)
- [x] [Redirect to the assignment override for a section](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.section_alias)
- [x] [Create an assignment override](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.create)
- [x] [Update an assignment override](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.update)
- [x] [Delete an assignment override](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.destroy)
- [x] [Batch retrieve overrides in a course](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.batch_retrieve)
- [x] [Batch create overrides in a course](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.batch_create)
- [x] [Batch update overrides in a course](https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.batch_update)

eelsirhc:
Can you give an example of the usage? This code doesn't work for me (assuming a valid assignment `va` and student `vs`
```
va.create_override(student_ids=[vs], due_at=datetime(2019,1,2,12,34))
```

instead returning an error
```
BadRequest: {"errors":["one of student_ids, group_id, or course_section_id is required"]}
```

If I manually construct the POST following the [Canvas documentation](https://github.com/instructure/canvas-lms/blob/master/app/controllers/assignment_overrides_controller.rb) I need to construct this

```
va.create_override(**{"assignment_override[student_ids]":[vs],
                                    "assignment_override[due_at]":datetime.datetime(2019,1,2,12,34)})
```

The keywords need the 'assignment_override' prefix, but you can't make them simple keywords anymore because of the square brackets in their name.

jessemcbride:
@eelsirhc Your call should look something like this:

```
override = assignment.create_override(
      assignment_override={
[...]

PR #217, opened 2018-11-07
Issue #216 Assignment Overrides

Thetwam:
Adds coverage for Assignment Override endpoints.

See #216

### DIFF
--- a/canvasapi/assignment.py
+++ b/canvasapi/assignment.py
@@ -328,6 +328,30 @@ class AssignmentOverride(CanvasObject):
     def __str__(self):
         return "{} ({})".format(self.title, self.id)
 
+    def delete(self, **kwargs):
+        """
+        Delete this assignment override.
+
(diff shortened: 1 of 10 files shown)


canvasapi/assignment.py
```diff
@@ -70,6 +88,54 @@ def get_gradeable_students(self, **kwargs):
...
+        )
+
+    def get_override(self, override, **kwargs):
+        """
+        Get the single override with the given override id.
+
+        :calls: `GET /api/v1/courses/:course_id/assignments/:assignment_id/overrides/:id
+        <https://canvas.instructure.com/doc/api/assignments.html#method.assignment_overrides.show>`_
+
+        :param override: The obejct or ID of the override to get
```
jessemcbride: "The obejct" -> "The object"

Also, above, in L97:

"Get the single override with the given override id." -> Get **a** single override with the...

Issue #254, opened 2019-06-07
Content Exports


Thetwam:
**What resource needs additional coverage?**
Content Exports

**What endpoints need to be covered?**

https://canvas.instructure.com/doc/api/content_exports.html

- [ ] List content exports
- [ ] Show content export
- [ ] Export content

Thetwam:
Resolved in #265

PR #259, opened 2019-06-12
Issue/254 content exports

weining-li:
- Added content exports API for course, group, and user.
 - Added unit test for content exports API.

Fixes issue #254


weining-li:
Oh wait I don't know why there are like 20 commits. I could start over if needed.

Thetwam:
> Oh wait I don't know why there are like 20 commits. I could start over if needed.

Not sure either. Will investigate.

PR #271, opened 2019-06-20
Quiz Submissions Paginated List

### LINKED ISSUE #270: quiz.get_all_quiz_submissions() Does not process all pages
# Describe the bug

quiz.get_all_quiz_submissions() only returns 10 submissions and does not handle multiple pages.

# To Reproduce

1) Create canvasapi client
2) Get a course
3) Get a quiz in that course that has more than 10 submissions
4) Call quiz.get_all_quiz_submissions() and you'll find that only 10 were returned

# Expected behavior

The page_size param should probably default to 100 and a pagination_list should be returned like other requests.

# Environment information (please complete the following information)

- Python version (`python --version`): 3.7.0
- CanvasAPI version (`pip show canvasapi`): 0.12.0

# Additional context

If the quiz is also an assignment, then assignment.get_submissions() works fine. However, if a quiz does not offer points then it is not also an assignment and you need to use quiz.get_all_quiz_submissions()

Thetwam:
# Proposed Changes

* Fixes an issue where `get_all_quiz_submissions` would only return up to the first 10 results
  * Switch from single request to `PaginatedList` with `_root` defined
* Deprecates `get_all_quiz_submissions` in favor of `get_submissions`
* Make `QuizSubmission` string representation more verbose

Fixes #270.


canvasapi/quiz.py
```diff
@@ -243,22 +245,43 @@ def get_all_quiz_submissions(self, **kwargs):
...
+        return self.get_submissions(**kwargs)
+
+    def get_submissions(self, **kwargs):
+        """
+        Get a list of all submissions for this quiz.
+
+        :calls: `GET /api/v1/courses/:course_id/quizzes/:quiz_id/submissions \
+        <https://canvas.instructure.com/doc/api/quiz_submissions.html#method.quizzes/quiz_submissions_api.index>`_
+
+        :rtype: list of :class:`canvasapi.quiz.QuizSubmission`
```
jessemcbride: rtype is PaginatedList, not list

Thetwam: Good catch. Changed.

PR #273, opened 2019-06-24
Issue #224 Generic HTTP Error Handling

### LINKED ISSUE #224: Implement 503 Handling
Upon receiving a 503 response from Canvas, the library returns a `json.decoder.JSONDecodeError` instead of something more useful.

Implement handling for 503 response.

It's probably worth implementing generic handling for all 4xx and 5xx responses.

Thetwam:
# Proposed Changes

* Add generic error handling for 4XX and above HTTP errors

Fixes #224.

### DIFF
--- a/canvasapi/requester.py
+++ b/canvasapi/requester.py
@@ -111,6 +111,11 @@ class Requester(object):
             raise Conflict(response.text)
         elif response.status_code == 500:
             raise CanvasException("API encountered an error processing your request")
+        elif response.status_code > 400:
+            # generic catch-all for error codes
+            raise CanvasException(
+                "Encountered an error: status code {}".format(response.status_code)
+            )
 
         return response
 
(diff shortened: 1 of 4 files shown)


canvasapi/requester.py
```diff
@@ -111,6 +111,11 @@ def request(
             raise Conflict(response.text)
         elif response.status_code == 500:
             raise CanvasException("API encountered an error processing your request")
+        elif response.status_code > 400:
```
jessemcbride: Is it redundant to check > 400 _and_ 500 status codes?

Thetwam: I had assumed we wanted a separate error message specifically for 500 since we had already specified one, but am open to merging the 500 and >400 responses

jessemcbride: I can't think of a reason to isolate 500 errors specifically, so I'm also okay with merging them together.

Issue #68, opened 2017-09-05
Formatting Standardization


jessemcbride:
Consider using EditorConfig to establish consistent style.

See PR #67 for discussion on this, as well as [EditorConfig's](http://editorconfig.org/) official site.

jessemcbride:
Following some internal discussion, this doesn't seem necessary. Running flake8 in our testing environment locally and with Travis seems sufficient.

jessemcbride:
@Thetwam I'd like to revisit this if we add black as a dependency for code formatting.

Thetwam:
Since @jessemcbride resurrected this issue, I'll just use it as a place to talk about standardizing formatting in general.

The library that Jesse mentioned (`black`) is a super cool auto-formatter:

- https://www.mattlayman.com/blog/2018/python-code-black/
- https://black.readthedocs.io/en/stable/

I've been playing with it over the last week and am really impressed. I would definitely like to use it in our project.

Due to the sheer volume of files it will touch, we'll have to time integrating it completely into the workflow such that we don't disrupt active development. In the meantime, I've been quietly using it on a case-by-case basis when reviewing PRs. 

I think I'll throw together a WIP PR to explore what integrating it into our workflow entails.

Thetwam:
@jessemcbride Now that we have `black` as part of our workflow, are you okay with re-closing this issue?

jessemcbride:
Yeah, `black` fixes this. Closing.

PR #278, opened 2019-07-01
[WIP] Format and Syntax Quality Control

Thetwam:
# Proposed Changes

*Note: This request is currently a work in progress*

**TODO:** Documentation for using `black` and mention the alphabetical script

* Add Black as a test dependency
* Add Black to TravisCI test runner
* Blacken all code
* Add test for methods to be in alphabetical order
* Alphabetize methods in `canvasapi` module

Related to (but does **not** resolve) #68


Thetwam:
`black` requires Python 3.6 or higher to run, so running it on older versions of Python (like we do with our Travis tests) isn't an option.

It does provide a way of targeting formatting of certain versions, though:

```
-t, --target-version [py27|py33|py34|py35|py36|py37|py38]
    Python versions that should be supported by
    Black's output. [default: per-file auto-detection]
```

PR #280, opened 2019-07-05
Grouped Submissions

### LINKED ISSUE #77: Grouped Responses
The `list_multiple_submissions` methods on `Course` and `Section` objects takes in an optional parameter `grouped` (See [Canvas Docs](https://canvas.instructure.com/doc/api/submissions.html#method.submissions_api.for_students)). If `grouped` is true (note: see #61 for details on boolean weirdness for this particular method), the format Canvas returns is different than what PaginatedList normally expects.

To get around this, we prevent the `grouped` parameter from being passed at all. Ideally, we should see if it's possible to parse this other format and leverage it's benefit of being grouped by user.

Similar issues may exist in other endpoints. It's possible that support for different formats is related to #56.

Thetwam:
# Proposed Changes

* Create a new `GroupedSubmission` class that wraps around multiple `Submission` objects.
* Change how `course.get_multiple_submissions` reacts when the `grouped` parameter is passed: Returns a `PaginatedList` of `GroupedSubmission` objects instead of throwing a warning and deleting the `grouped` parameter.
* Formatting tweaks

Fixes #77.

### DIFF
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -18,7 +18,7 @@ from canvasapi.paginated_list import PaginatedList
 from canvasapi.progress import Progress
 from canvasapi.quiz import QuizExtension
 from canvasapi.tab import Tab
-from canvasapi.submission import Submission
+from canvasapi.submission import GroupedSubmission, Submission
 from canvasapi.upload import Uploader
 from canvasapi.util import combine_kwargs, is_multivalued, file_or_path, obj_or_id
 from canvasapi.rubric import Rubric
@@ -1411,8 +1411,14 @@ class Course(CanvasObject):
             :class:`canvasapi.submission.Submission`
         """
         if 'grouped' in kwargs:
-            warnings.warn('The `grouped` parameter must be empty. Removing kwarg `grouped`.')
-            del kwargs['grouped']
+            return PaginatedList(
+                GroupedSubmission,
+                self._requester,
+                'GET',
+                'courses/{}/students/submissions'.format(self.id),
+                {'course_id': self.id},
+                _kwargs=combine_kwargs(**kwargs)
+            )
 
         return PaginatedList(
             Submission,
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -1413,7 +1413,9 @@ class Course(CanvasObject):
             :class:`canvasapi.submission.Submission`
         """
 
-        if "grouped" in kwargs and normalize_bool(kwargs["grouped"], "grouped"):
+        is_grouped = kwargs.get("grouped", False)
+
+        if normalize_bool(is_grouped, "grouped"):
             cls = GroupedSubmission
         else:
             cls = Submission
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -1410,23 +1410,15 @@ class Course(CanvasObject):
         :rtype: :class:`canvasapi.paginated_list.PaginatedList` of
             :class:`canvasapi.submission.Submission`
         """
-        if 'grouped' in kwargs:
-            return PaginatedList(
-                GroupedSubmission,
-                self._requester,
-                'GET',
-                'courses/{}/students/submissions'.format(self.id),
-                {'course_id': self.id},
-                _kwargs=combine_kwargs(**kwargs)
-            )
+        cls = GroupedSubmission if 'grouped' in kwargs else Submission
 
         return PaginatedList(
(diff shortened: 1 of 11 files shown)


canvasapi/submission.py
```diff
@@ -188,3 +188,20 @@ def upload_comment(self, file, **kwargs):
...
+
+
+@python_2_unicode_compatible
+class GroupedSubmission(CanvasObject):
+
+    def __str__(self):
+        return "{} submission(s) for User #{}".format(len(self.submissions), self.user_id)
+
+    def __init__(self, requester, attributes):
+        self.submissions = list()
```
jessemcbride: Alternate possibility, not sure if I like this more:

```
try:
    self.submissions = [Submission(requester, submission) for submission in attributes['submissions']]
    del attributes['submissions']
except KeyError:
     pass
```

canvasapi/course.py
```diff
@@ -1411,8 +1411,14 @@ def get_multiple_submissions(self, **kwargs):
             :class:`canvasapi.submission.Submission`
         """
         if 'grouped' in kwargs:
-            warnings.warn('The `grouped` parameter must be empty. Removing kwarg `grouped`.')
-            del kwargs['grouped']
+            return PaginatedList(
```
jessemcbride: Instead of having separate returns, what about this?

```
cls = Submission

if 'grouped' in kwargs:
    cls = GroupedSubmission

return PaginatedList(
     cls,
     self._requester,
     'GET',
     'courses/{}/students/submissions'.format(self.id),
     {'course_id': self.id},
    _kwargs=combine_kwargs(**kwargs)
)
```

canvasapi/course.py
```diff
@@ -1410,17 +1410,15 @@ def get_multiple_submissions(self, **kwargs):
         :rtype: :class:`canvasapi.paginated_list.PaginatedList` of
             :class:`canvasapi.submission.Submission`
         """
-        if 'grouped' in kwargs:
-            warnings.warn('The `grouped` parameter must be empty. Removing kwarg `grouped`.')
-            del kwargs['grouped']
+        cls = GroupedSubmission if 'grouped' in kwargs else Submission
```
jessemcbride: This will produce a bug if someone passed `grouped=False` to `get_multiple_submissions`.

Let's change this to:

```
cls = GroupedSubmission if kwargs.get('grouped') else Submission
```

tests/test_course.py
```diff
@@ -918,23 +918,13 @@ def test_get_multiple_submissions(self, m):
...
-            submission_list = [submission for submission in submissions]
-
-            # Ensure using the `grouped` param raises a warning
-            self.assertEqual(len(warning_list), 1)
-            self.assertEqual(warning_list[-1].category, UserWarning)
-            self.assertEqual(
-                text_type(warning_list[-1].message),
-                'The `grouped` parameter must be empty. Removing kwarg `grouped`.'
-            )
+        submissions = self.course.get_multiple_submissions(grouped=True)
```
jessemcbride: We should write a test to ensure that `grouped=False` behaves correctly.

canvasapi/course.py
```diff
@@ -1410,17 +1412,19 @@ def get_multiple_submissions(self, **kwargs):
         :rtype: :class:`canvasapi.paginated_list.PaginatedList` of
             :class:`canvasapi.submission.Submission`
         """
-        if 'grouped' in kwargs:
-            warnings.warn('The `grouped` parameter must be empty. Removing kwarg `grouped`.')
-            del kwargs['grouped']
+
+        if "grouped" in kwargs and normalize_bool(kwargs["grouped"], "grouped"):
```
jessemcbride: This might be more easily read as:

```
grouped = kwargs.get("grouped")

if normalize_bool(grouped, "grouped"):
    cls = GroupedSubmission
else:
    cls = Submission
```

Thetwam: That is not logically equivalent. If there is no `grouped` value passed, then `kwargs.get("grouped")` will return `None`. 

Then, `normalize_bool(grouped, "grouped")` will throw an exception.

jessemcbride: Two options, then:

`grouped = kwargs.get("grouped", False)` OR

```
grouped = kwargs.get("grouped")

try:
    if normalize_bool(grouped, "grouped"):
        cls = GroupedSubmission
    else:
         cls = Submission
except ValueError:
    cls = Submission
```

PR #339, opened 2019-11-15
Update iterable check

### LINKED ISSUE #284: importing ABCs from collections will be deprecated in Python 3.8
I get this warning every time I import canvasapi:
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

Thetwam:
Use `iter()` instead of checking if instance `Iterable`. See https://github.com/ucfopen/canvasapi/issues/284#issuecomment-533184311 for explanation.

Fixes #284.

### DIFF
--- a/canvasapi/util.py
+++ b/canvasapi/util.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import os
-from collections import Iterable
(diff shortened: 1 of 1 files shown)


jessemcbride:
LGTM

Issue #341, opened 2019-11-18
Deprecate `attributes` attribute


Thetwam:
`CanvasObject.set_attributes()` sets an attribute called `attributes` that includes all the attributes (as a `dict`) sent to the function. This is not used anywhere in the function, or anywhere else in the library except for the `to_json` function.

Based on its docstring, the intention of `to_json` is to return the original JSON that created the object. As we've discovered (see #303), this is not actually the case. `self.attributes` can be created, mutated, and updated in many different ways that `to_json` does not accurately reflect. This information can be gathered from `Requester._cache` much more reliably. I also have a suspicion that most people don't really use `to_json` anyways.

We should deprecate `attributes` and `to_json()`, and ultimately remove them in a later release.

Thetwam:
Related: #387

PR #342, opened 2019-11-18
Deprecate CanvasObject.attributes and CanvasOject.to_json()

Thetwam:
Related: #341


canvasapi/canvas_object.py
```diff
@@ -77,5 +87,14 @@ def to_json(self):
...
         Return the original JSON response from the API that was used to
         construct the object.
+
+        .. warning::
+            .. deprecated:: 0.14.0
+                Use :func:`canvasapi.assignment.Assignment.get_submission` instead.
         """
+        warnings.warn(
+            "`CanvasObject.to_json()` is deprecated and will be removed in a future version. "
+            "To see the original responses, check Requester._cache",
```
jessemcbride: Should we tell them to enable request logging instead?

Thetwam: That sounds good! Could you please provide a better warning string?

jessemcbride:
```
warnings.warn(
 "`CanvasObject.to_json()` is deprecated and will be removed in a future version. "
 "To view the original attributes sent by Canvas, enable logs from the requests library",
)
```

canvasapi/canvas_object.py
```diff
@@ -77,5 +87,15 @@ def to_json(self):
         """
         Return the original JSON response from the API that was used to
         construct the object.
+
+        .. warning::
+            .. deprecated:: 0.14.0
+                Use :func:`canvasapi.assignment.Assignment.get_submission` instead.
```
jessemcbride: We need to update this text.

Thetwam: Whoops! Good catch

canvasapi/canvas_object.py
```diff
@@ -77,5 +87,16 @@ def to_json(self):
         """
         Return the original JSON response from the API that was used to
         construct the object.
+
+        .. warning::
+            .. deprecated:: 0.14.0
```
jessemcbride: Deprecation is actually happening in 0.15.0, right?

Thetwam: True! Fixed.

PR #355, opened 2020-02-13
issue/354 graphql support

### LINKED ISSUE #354: Add support for GraphQL endpoint
[GraphQL](https://canvas.instructure.com/doc/api/file.graphql.html) is a new endpoint that would be nice if it had a way to query it via this API. 

Initial implementation would just be to support this endpoint `POST /api/graphql` for making raw JSON POST requests and getting back a dict response. Future enhancements might be to switch over existing calls to optionally use this endpoint rather than going through the Canvas API methods for improved performance. 

I'm working on a proof or concept that I could use some review on.

jonespm:
# Proposed Changes

* Add support for the Canvas GraphQL Endpoint
* Make the API version optional defaulting to v1 (GraphQL doesn't currently use an API version and it's possible there may be future API's on v2, v3, etc.

Feedback is appreciated. Most of the difficult changes here were around the v1 being hardcoded and needing to pass the query as one JSON string rather than form parameters.

Fixes #354

### DIFF
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -1182,18 +1182,12 @@ class Canvas(object):
             _kwargs=combine_kwargs(**kwargs),
         )
 
-    def graphql(self, query, variables):
+    def graphql(self, **kwargs):
         """
         Makes a GraphQL formatted requeset to Canvas
 
         :calls: `POST /api/graphql \
         <https://canvas.instructure.com/doc/api/file.graphql.html>`_
-
-        :param query: The GraphQL query to execute
-        :type query: str
-
-        :param variables: Variable values as required by the supplied query
-        :type variables: dict
 
         :rtype: dict
         """
@@ -1201,7 +1195,7 @@ class Canvas(object):
             "POST",
             "graphql",
             headers={"Content-Type": "application/json"},
-            _kwargs=[("query", query), ("variables", variables)],
+            _kwargs=combine_kwargs(flatten=False, **kwargs),
             # Needs to call special endpoint without api/v1
             _url=self.__requester.original_url + "/api/graphql",
             json=True,
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import warnings
+import json
 
 from canvasapi.account import Account
 from canvasapi.course import Course
@@ -64,7 +65,7 @@ class Canvas(object):
         # the API.
         access_token = access_token.strip()
 
-        base_url = new_url + "/api/v1/"
+        base_url = new_url + "/api/"
 
         self.__requester = Requester(base_url, access_token)
 
@@ -1382,3 +1383,27 @@ class Canvas(object):
             "PUT", "users/self/course_nicknames/{}".format(course_id), nickname=nickname
         )
         return CourseNickname(self.__requester, response.json())
+
+    def graphql(self, query, variables):
+        """
+        Makes a GraphQL formatted requeset to Canvas
+
+        :calls: `POST /api/graphql \
+        <https://canvas.instructure.com/doc/api/file.graphql.html>`_
+
+        :param query: The GraphQL query to execute
+        :type query: str
+
+        :param variables: Variable values as required by the supplied query
+        :type variables: dict
+
+        :rtype: dict
+        """
+        response = self.__requester.request(
+            "POST",
+            "graphql",
+            _apiv = '',
+            headers = {"Content-Type": "application/json"},
+            _kwargs= json.dumps({"query":query, "variables":variables})
+        )
+        return response.json()
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -1184,7 +1184,7 @@ class Canvas(object):
 
     def graphql(self, query, variables=None, **kwargs):
         """
-        Makes a GraphQL formatted requeset to Canvas
+        Makes a GraphQL formatted request to Canvas
 
         :calls: `POST /api/graphql \
         <https://canvas.instructure.com/doc/api/file.graphql.html>`_
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -1182,12 +1182,17 @@ class Canvas(object):
             _kwargs=combine_kwargs(**kwargs),
         )
 
-    def graphql(self, **kwargs):
+    def graphql(self, query, variables=None, **kwargs):
         """
         Makes a GraphQL formatted requeset to Canvas
 
         :calls: `POST /api/graphql \
         <https://canvas.instructure.com/doc/api/file.graphql.html>`_
+
+        :param query: The GraphQL query to execute as a String
+        :type query: str
+        :param variables: The variable values as required by the supplied query
+        :type variables: dict
 
         :rtype: dict
         """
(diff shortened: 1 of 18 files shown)


jonespm:
I'll fix up these flake8 issues

canvasapi/requester.py
```diff
@@ -90,16 +91,17 @@ def _post_request(self, url, headers, data=None):
...
-            if field == "file":
-                if isinstance(value, dict):
-                    files = value
-                else:
-                    files = {"file": value}
-                break
-
-        # Remove file entry from data.
-        data[:] = [tup for tup in data if tup[0] != "file"]
+        if not isinstance(data, str):
```
jessemcbride: Could you walk me through this change?

jonespm: So the idea here is that you could send data as a just a pre-processed string or as a dict. I've used the canvasapi and know more about it so let me look to see if there's a better way of doing this.

I feel like the biggest problems were more around getting the hardcoded "apiv1" out of this and getting all the tests to pass. This change and the other one around `if hasattr(_kwargs, "extend"): ` that was meant to ensure the value was also a dict might not be necessary. I'll try to get to re-review it this week.

jonespm: Okay so now I remember what I was doing here.
It looks like for a general query to Canvas you're using [this Tuple pattern for requests and passing it as form style data](https://requests.readthedocs.io/en/master/user/quickstart/#more-complicated-post-requests
).

However, the GraphQL api takes json data that's passed in as json. I figured the *easiest* way to get it into requester was to just pass the string already created as json and basically have the requester.py not do any processing to it. (This is what's in this PR)

The alternate would be to keep the tuples and have the POST (and PUT) methods will need to do the conversion from tuples to dict and do the json.dumps there before making the call (or use the `json` parameter of requests.post). I could do this change, it might be less code changes but would probably require passing along another kwarg.

jonespm: Okay, I redid this code to be option #2 and it's still working as expected.

jonespm:
I modifed this PR to still pass through the tuples but pass it as a json as the last step. I'm only able to get the graphql api on Canvas to work if I pass everything through as json rather than the regular data. I added `kwargs` to all of the methods to avoid having to do a special case for POST and to possibly support PUT and PATCH if Canvas ever does support those methods on this endpoint.

canvasapi/requester.py
```diff
@@ -173,7 +189,7 @@ def request(
             elif isinstance(arg, datetime):
                 _kwargs[i] = (kw, arg.isoformat())
 
-        # Determine the appropriate request method.
+            # Determine the appropriate request method.
```
jonespm, line 192:
```suggestion
        # Determine the appropriate request method.
```

canvasapi/canvas.py
```diff
@@ -1184,6 +1184,32 @@ def get_user_participants(self, appointment_group, **kwargs):
...
+        :type variables: dict
+
+        :rtype: dict
+        """
+        response = self.__requester.request(
+            "POST",
+            "graphql",
+            _apiv="",
+            headers={"Content-Type": "application/json"},
+            _kwargs=[("query", query), ("variables", variables)],
```
jessemcbride, line 1207: Rather than constructing kwargs manually, I think using `canvasapi.util.combine_kwargs` here might make this more consistent with the rest of the library:

```python
    def graphql(self, **kwargs):
        """
        Makes a GraphQL formatted request to Canvas.

        :calls: `POST /api/graphql <https://canvas.instructure.com/doc/api/file.graphql.html>`_

        :rtype: dict
        """
        response = self.__requester.request(
            "POST",
            _url=self.__requester.base_url + "api/graphql",
            headers={"Content-Type": "application/json"},
            json=True,
        )

        return response.json()
```

This _should_ enable us to call this with kwargs directly:
```python
canvas.graphql(query=
   """
   query courseInfo($courseId: ID!) {
       course(id: $courseId) {
        id
        _id
        name
       }
     }
    """, variables={"courseId": 1})
```

I'm not sure what implications sending this as JSON instead of form data will have, but historically we've sent parameters like `variables[courseId]=1` using the combine_kwargs tuple recipe `variables={"courseId": 1}`. Something to think about.

canvasapi/requester.py
```diff
@@ -60,10 +61,12 @@ def _get_request(self, url, headers, params=None):
         :type headers: dict
         :param params: The parameters to send with this request.
         :type params: dict
+        :param json: Whether or not to send the data as json
```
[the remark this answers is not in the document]
jonespm: Yeah probably was based on an older change, removing.

canvasapi/canvas.py
```diff
@@ -1184,6 +1184,32 @@ def get_user_participants(self, appointment_group, **kwargs):
...
+
+        :param variables: Variable values as required by the supplied query
+        :type variables: dict
+
+        :rtype: dict
+        """
+        response = self.__requester.request(
+            "POST",
+            "graphql",
+            _apiv="",
```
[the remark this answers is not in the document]
jonespm: Yeah, I didn't see this before. I mean it seems a little hacky but I also haven't seen anything other than /api/v1 and this /api for GraphQL. Instructure says all new functionality is going to get added to graphQL. I'd figure if they did significantly change something they'd go v2 (as that's the point of v1) which is why I added these changed but as you say this doesn't look required to accomplish this! I noticed the `_url `when working on another recent fix.

I can revert these changes and see what it looks like, I feel like this should be fine.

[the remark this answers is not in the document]
jonespm: Ok, this seems close but the problem I'm getting now is `__requester.base_url `gets set to be "/api/v1" in the Canvas `__init__ `. So a little more manipulation might be needed, or we'll have to store the original URL?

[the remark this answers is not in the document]
jonespm: I changed a little around here, appending the "/api/v1" in the requester instead of passing it into Canvas, then preserving the original URL. Let me know what you think about this.

canvasapi/canvas.py
jonespm: Ok I can try this. There was only two args query and variables that were supported by this endpoint.

jonespm: When I'm running this through combine_kwargs it's making the dict
` ('variables[termid]', 125)`
for this endpoint it seems to needs to stay as

`('variables', {'termid': 125})`

We'd need some way to stop combined_kwargs from doing anything with a dict. I can commit something maybe you have a better idea?

Thetwam: If we can guarantee that `query` and `variables` are the only params, I don't mind constructing `_kwargs` manually. `combine_kwargs` is gonna screw with a lot of things people could potentially pass in.

This endpoint is fundamentally different from most of our endpoints, so I don't mind being a little inconsistent here.

jonespm: Yeah, it's either we
* Have fixed args here and construct it manually
* Use kwargs but pop them out and construct it manually
* Change combine_kwargs to avoid messing with this (I did a commit for this but am happy to switch it out as it seems hacky and only for this)

Thetwam: `query` and `variables` as fixed args, with a catch-all `**kwargs` at the end makes the most sense to me.

Would want to run `combine_kwargs` against `kwargs` first, then add `query` and `variables` to it before passing to `_kwargs`.

I'm not a fan of 3. Like you said, it seems hacky.

jonespm: Okay, I made this change

canvasapi/requester.py
```diff
@@ -124,6 +131,7 @@ def request(
         use_auth=True,
         _url=None,
         _kwargs=None,
+        json=False,
```
jessemcbride, line 134: Is hardcoding `json=False` here intentional?

canvasapi/canvas.py
```diff
@@ -1184,6 +1182,33 @@ def get_user_participants(self, appointment_group, **kwargs):
             _kwargs=combine_kwargs(**kwargs),
         )
 
+    def graphql(self, query, variables=None, **kwargs):
+        """
+        Makes a GraphQL formatted requeset to Canvas
```
jessemcbride, line 1187: Tiny typo in `request` in the docstring (`requeset`).

PR #361, opened 2020-04-09
Pass report_type param correctly in create_report

brucespang:
Hi! It seems like creating a report on a quiz currently does not work. The Canvas api [expects](https://github.com/instructure/canvas-lms/blob/master/app/controllers/quizzes/quiz_reports_controller.rb#L159) the report type to be passed as `quiz_report[report_type]`, but canvasapi passes it as `report_type`. This causes an error like:
```
~/code/canvasapi/canvasapi/requester.py in request(self, method, endpoint, headers, use_auth, _url, _kwargs, **kwargs)
    220         # Raise for status codes
    221         if response.status_code == 400:
--> 222             raise BadRequest(response.text)
    223         elif response.status_code == 401:
    224             if "WWW-Authenticate" in response.headers:

BadRequest: {"status":"bad_request","message":"invalid quiz report type","error_report_id":53209570}
```

This change attempts to fix that error by passing the report_type in the correct parameter. I've tested it locally and it seems to work

### DIFF
--- a/canvasapi/quiz.py
+++ b/canvasapi/quiz.py
@@ -143,7 +143,7 @@ class Quiz(CanvasObject):
                 "Param `report_type` must be a either 'student_analysis' or 'item_analysis'"
             )
 
-        kwargs["report_type"] = report_type
+        kwargs["quiz_report"] = {"report_type": report_type}
 
         response = self._requester.request(
             "POST",


jonespm:
This does look correct now based on the API. https://canvas.instructure.com/doc/api/quiz_reports.html#method.quizzes/quiz_reports.create

The alternative (I believe) would have been to remove the arg `report_type` and just prefer to call this method as

`create_report(quiz_report={'report_type': 'student_analysis'})`

jessemcbride:
Hey @brucespang - thanks for the patch! We should be able to get this in our next release.

PR #388, opened 2020-07-06
Remove deprecated methods and attributes

### LINKED ISSUE #341: Deprecate `attributes` attribute

### LINKED ISSUE #387: Remove Deprecated Methods
We have several methods throughout the library marked a deprecated. Remove them completely.

Thetwam:
# Proposed Changes

* Remove deprecated methods and attributes.

Fixes #387.
Fixes #341.


canvasapi/canvas.py
```diff
@@ -60,7 +56,7 @@ def __init__(self, base_url, access_token):
         # Ensure that the user-supplied access token and base_url contain no leading or
         # trailing spaces that might cause issues when communicating with the API.
         access_token = access_token.strip()
-        base_url = new_url.strip()
+        base_url = get_institution_url(base_url)
```
jessemcbride, line 59: Good catch.

PR #398, opened 2020-07-14
Fix Quiz Group Course ID

hcolclou:
# Proposed Changes

Previously, Quiz.get_quiz_group set the course ID of the quiz group to be the ID of the quiz. This caused errors when using the constructed QuizGroup to make requests as the course ID was incorrect causing the URL to be incorrect. This change fixes that issue by setting the course ID of the quiz group to be the course ID of the course the quiz is contained in.


Thetwam:
Hey @hcolclou,

Thanks for the pull request. Great catch! 👍 

I also altered the test for this function (and related ones) to prevent this from happening in the future.

PR #403, opened 2020-07-23
Issue #402 Create External Tool - Client ID

### LINKED ISSUE #402: Create an External Tool LTI 1.3 Support
`Course.create_external_tool()` has four required parameters: `name`, `privacy_level`, `consumer_key`, and `shared_secret`. 

When Instructure added LTI 1.3 support, they added an additional "required" parameter called `client_id`. Although all five of these parameters are marked as "required" in the documentation, the actual requirement is to use either `client_id` **_or_** the other four params.

We need to update `create_external_tool()` to accept `client_id` without requiring the other four parameters.

Note that this will be a breaking change for anyone using positional arguments in `create_external_tool()`, but should work without issue for those already using keyword args.

Thetwam:
# Proposed Changes

* Remove positional arguments from `Course.create_external_tool` to allow `client_id` to be passed for LTI 1.3 support.

Fixes #402.

### DIFF
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -275,9 +275,7 @@ class Course(CanvasObject):
         )
         return ExternalFeed(self._requester, response.json())
 
-    def create_external_tool(
-        self, name, privacy_level, consumer_key, shared_secret, **kwargs
-    ):
+    def create_external_tool(self, **kwargs):
         """
         Create an external tool in the current course.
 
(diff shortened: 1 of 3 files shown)


canvasapi/course.py
```diff
@@ -286,24 +284,21 @@ def create_external_tool(
...
-        :param consumer_key: The consumer key for the external tool
-        :type consumer_key: str
-        :param shared_secret: The shared secret with the external tool
-        :type shared_secret: str
+
         :rtype: :class:`canvasapi.external_tool.ExternalTool`
         """
         from canvasapi.external_tool import ExternalTool
 
+        required_params = ("name", "privacy_level", "consumer_key", "shared_secret")
```
jessemcbride, line 292: I wonder if we could write a utility to wrap our `RequiredFieldMissing` code. Definitely outside the scope of this patch, but it seems like I've seen this pattern before. Food for thought.

Thetwam: Not a bad idea. We handle it slightly differently in a lot of places, might be useful to standardize.

PR #417, opened 2020-09-08
Fixup broken links to readthedocs in README

jonespm:
# Proposed Changes

* URL changed from latest to stable


jessemcbride:
What a pal!

Good catch. Thank you :)

jonespm:
Yeah I didn't think it'd get merged this fast! :) The only other place I see this referenced is in the CHANGELOG but those are for old changes and probably not worth updating. I'm not sure why it changed from latest->stable.

PR #421, opened 2020-10-01
Issue/420 fetch announcements

### LINKED ISSUE #420: fetching announcements from a `Canvas` object doesn't assign the `course_id`
# Describe the bug

Retrieving announcements from Canvas happen at the level of the `Canvas` object, not course level.  Getting an announcement and trying to call `a.update()` fails generating this error:
```
ValueError: Discussion Topic does not have a course_id or group_id
```

A workaround is to manually assign the course_id into the announcement, and then update.  The following succeeds:
```
b.course_id = 123456
b.update()
```

 

# To Reproduce

```
import canvasapi
canvas = canvasapi.Canvas(API_URL, API_KEY)

course_id = 123456
announcements = canvas.get_announcements(context_codes=['course_{}'.format(course_id)])
a = announcements[0]
a.update()
```

This is the complete error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-15-667b0a4b45f2> in <module>
      2 announcements = canvas.get_announcements(context_codes=['course_{}'.format(course_id)])
      3 a = announcements[0]
----> 4 a.update()

~/opt/anaconda3/lib/python3.7/site-packages/canvasapi/discussion_topic.py in update(self, **kwargs)
    309             "PUT",
    310             "{}s/{}/discussion_topics/{}".format(
--> 311                 self._parent_type, self._parent_id, self.id
    312             ),
    313             _kwargs=combine_kwargs(**kwargs),

~/opt/anaconda3/lib/python3.7/site-packages/canvasapi/canvas_object.py in __getattribute__(self, name)
     15 
     16     def __getattribute__(self, name):
---> 17         return super(CanvasObject, self).__getattribute__(name)
     18 
     19     def __init__(self, requester, attributes):

~/opt/anaconda3/lib/python3.7/site-packages/canvasapi/discussion_topic.py in _parent_type(self)
     34             return "group"
     35         else:
---> 36             raise ValueError("Discussion Topic does not have a course_id or group_id")
     37 
     38     def delete(self, **kwargs):

ValueError: Discussion Topic does not have a course_id or group_id
```
# Expected behavior

I expect the line
```
a.update()
```
to run cleanly,  and without error.

# Environment information

- Python 3.7.4
- CanvasAPI version 2.0.0

# Additional context

My ultimate goal is to be able to programmatically adjust `delayed_post_at` for announcements.

bennettscience:
# Proposed Changes

* Add a check for `context_code` on `DiscussionTopic` objects to assign a course ID for announcements.

Fixes #420


.gitignore
```diff
@@ -21,3 +21,4 @@
 setup.sh
 .python-version
 .*.swp
+venv/
```
Thetwam, line 24: I'm fine with adding `/venv/` to the ignore, but please move it near the other virtual environment dirs (`/env/`, `/env3/`), and ensure there's a newline at the end of the file.

canvasapi/discussion_topic.py
```diff
@@ -16,6 +16,9 @@ def _parent_id(self):
         """
         if hasattr(self, "course_id"):
             return self.course_id
+        elif hasattr(self, "context_code"):
+            course_id = self.context_code.split("_")[1]
+            return course_id
```
Thetwam, line 21: I believe it's possible for the context code to have values other than `course_x`, such as `group_x`.

I think this should also come after the `group_id` check, as the last check before the `else`.

Also, please add test coverage for this section.

canvasapi/discussion_topic.py
```diff
@@ -30,6 +33,8 @@ def _parent_type(self):
         """
         if hasattr(self, "course_id"):
             return "course"
+        elif hasattr(self, "context_code"):
+            return "course"
```
Thetwam, line 37: As noted above, I think it's possible for the `context_code` to be of type other than `"course"`.

Thetwam (changes requested):
Looks like a solid approach, just needs some refinement.

canvasapi/discussion_topic.py
bennettscience: I went through the docs for announcements, groups, and discussion topics and only announcements had the `context_code` key and it specifically states that only courses are supported right now. I don't think it's a bad idea to add a check...are you okay with a simple `or` in the check?

Thetwam: Fair enough!

So long as you're sure there's no way for it to have a `group_id` and `context_code`, then the `or` is fine.

bennettscience:
OK, here's another question:

This call requires the `context_codes[]` parameter. Calling `canvas.get_announcements()` fails if the user hasn't defined a `course_12345` string. Is it better to document this as a required `**kwarg` or add a required param to the method? There are other optional arguments that can be passed, so I'm tempted to make it a required parameter for the call, but I'll do whatever you think is best.


```python
def get_announcements(self, course_id, **kwargs):
        """
        List announcements.

        :calls: `GET /api/v1/announcements \
        <https://canvas.instructure.com/doc/api/announcements.html#method.announcements_api.index>`_

        :param course_id: Course ID to request announcements from.

        :rtype: :class:`canvasapi.paginated_list.PaginatedList` of
                :class:`canvasapi.discussion_topic.DiscussionTopic`
        """
        from canvasapi.discussion_topic import DiscussionTopic

        return PaginatedList(
            DiscussionTopic,
            self.__requester,
            "GET",
            "announcements",
            context_codes=[f'course_{course_id}'],
        )
```

bennettscience:
Not sure why the build failed...is there any way to get a log or report of what happened?

Thetwam:
@bennettscience looks like the `black` check failed:

https://travis-ci.org/github/ucfopen/canvasapi/builds/732295563

Thetwam:
> This call requires the `context_codes[]` parameter. Calling `canvas.get_announcements()` fails if the user hasn't defined a `course_12345` string. Is it better to document this as a required `**kwarg` or add a required param to the method? There are other optional arguments that can be passed, so I'm tempted to make it a required parameter for the call, but I'll do whatever you think is best.

In general, we try to make required parameters into required arguments for the function (not sure how we missed this argument previously! 😅).

Since `context_codes[]` is required by Canvas, I'd say add a required positional parameter called `context_codes` of type `list`.

However, this would require the user to build their own context string which, while not difficult, is mildly annoying. I like your approach of taking in a list of course IDs instead (especially if we leverage `util.obj_or_id()` and accept `Course` objects as well).

I'm honestly torn on which solution is better. Just making `context_codes` required with no additional processing would be more consistent with what we usually do, but we'd lose out on extra usability. Alternatively, adding the extra functionality would be super useful, but potentially confusing for those users who expect to be able to pass a list of `course_xxxxx` strings. At the risk of overcomplicating things, maybe there's a hybrid option where we can have both work?

Thoughts? @jessemcbride @bennettscience

canvasapi/discussion_topic.py
```diff
@@ -18,6 +20,13 @@ def _parent_id(self):
             return self.course_id
         elif hasattr(self, "group_id"):
             return self.group_id
+        elif hasattr(self, "context_code"):
+            if re.match("course", self.context_code):
```
Thetwam, line 24: If the intention here is to parse the string to figure out if it contains `"course"` (or later, `"group"`), that can be achieved without `re` with `"course" in self.context_code`.

If position is important, `self.context_code.startswith("course")` should also work.

.gitignore
```diff
@@ -15,9 +15,10 @@
...
 /env/
 /env3/
+venv/
 /htmlcov/
 \#*#
 .vscode/settings.json
 setup.sh
 .python-version
-.*.swp
+.*.swp
```
Thetwam, line 24: Please add a newline to the end of the file

canvasapi/canvas.py
```diff
@@ -437,23 +437,32 @@ def get_activity_stream_summary(self, **kwargs):
...
+            raise RequiredFieldMissing("Course IDs need to be passed as a list")
+
+        course_ids = [f"course_{course_id}" for course_id in course_list]
+
         return PaginatedList(
             DiscussionTopic,
             self.__requester,
             "GET",
             "announcements",
+            context_codes=course_ids,
```
Thetwam, line 465: Discussion for how to handle `context_codes` is ongoing, but regardless of what decision we come to, `context_codes` will almost certainly be a `list`. For our parameter "unpacking"/"combining" to work, it will need to be run through `combine_kwargs`.

The best way to do that is to add `context_codes` into `kwargs` before the `combine_kwargs` call.

Something like this, before the PaginatedList object:
```python
kwargs['context_codes'] = course_ids
```

Thetwam (changes requested):
I think we're on the right track. Had some thoughts about how to handle the `context_code` parameter above.

canvasapi/discussion_topic.py
bennettscience: Wouldn't checking for "course" using `re` throw a `KeyError` if the context code isn't present? That's why I checked with `hasattr` first.

Thetwam: Yeah checking with `hasattr` first is fine. I'm saying you don't need to use `re`:

```suggestion
            if self.context_code.startswith("course"):
```

jessemcbride:
> I'm honestly torn on which solution is better. Just making `context_codes` required with no additional processing would be more consistent with what we usually do, but we'd lose out on extra usability. Alternatively, adding the extra functionality would be super useful, but potentially confusing for those users who expect to be able to pass a list of `course_xxxxx` strings. At the risk of overcomplicating things, maybe there's a hybrid option where we can have both work?

I think a hybrid approach would be alright here in the spirit of more usability. If Canvas ever changes the endpoint (to allow other context code types, for example), we can rethink the approach, but for now, I think this works:

* Take in a list of one or more `courses` as a required argument.
* Check to see if courses is a list of context codes (by seeing if the first element in `courses` is a string containing `course_`). If it is, skip the `obj_or_id` branch.
* If it isn't, use `obj_or_id` to enable users to pass a Course object or an id for convenience.
* Format the list of context_codes appropriately.
* Send off the request.

bennettscience:
Sorry this PR was such a saga. I need to add automatic formatting on save...

jessemcbride (approved):

Thetwam (approved):

Thetwam:
@bennettscience Great work, thanks for the PR! 👍

PR #422, opened 2020-10-01
Solved Issue #367 CurrentUser Documentation

### LINKED ISSUE #367: CurrentUser Documentation
The function `get_current_user` in `canvas.py` doesn't have an entry in our docs. I suspect this is because it doesn't have a docstring.

Additionally, it isn't obvious from our documentation that `CurrentUser` objects have all the methods of the `User` class.

- [ ] Add a docstring to `get_current_user`, including description and return type.
- [ ] Add text to `current-user-ref.rst` explaining that the `CurrentUser` class is a subclass of the `User` class, and thus inherits all of the methods of `User`.
- [ ] In the explanation above, provide a link to the `User` class reference page.

Xx-Ashutosh-xX:
# Proposed Changes
* Added Docstring to get_user_method in canvas.py
* Added some Documentation in current-user-ref.rst explaining the inheritance of current_user class from the User class
* Added the reference to the User class 
Fixes #367

### DIFF
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -744,9 +744,6 @@ class Canvas(object):
         :calls: `GET /api/v1/users/:user_id \
         <https://canvas.instructure.com/doc/api/users.html#method.current_user.show>`_
 
-        :param user: The object or ID of the current user.
-        :type user: :class:`cavasapi.user.User` or int
-
         :rtype: :class:`canvasapi.current_user.CurrentUser`
         """
         return CurrentUser(self.__requester)
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -738,6 +738,17 @@ class Canvas(object):
         )
 
     def get_current_user(self):
+        """
+        Return a details of the current user.
+
+        :calls: `GET /api/v1/users/:user_id \
+        <https://canvas.instructure.com/doc/api/users.html#method.current_user.show>`_
+
+        :param user: The object or ID of the current user.
+        :type user: :class:`cavasapi.user.User` or int
+
+        :rtype: :class:`canvasapi.current_user.CurrentUser`
+        """
         return CurrentUser(self.__requester)
 
     def get_epub_exports(self, **kwargs):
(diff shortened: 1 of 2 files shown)


canvasapi/canvas.py
```diff
@@ -738,6 +738,17 @@ def get_courses(self, **kwargs):
...
 
     def get_current_user(self):
+        """
+        Return a details of the current user.
+
+        :calls: `GET /api/v1/users/:user_id \
+        <https://canvas.instructure.com/doc/api/users.html#method.current_user.show>`_
+
+        :param user: The object or ID of the current user.
+        :type user: :class:`cavasapi.user.User` or int
```
Thetwam, line 748: `:param:` and `:type:` are used to document the arguments of the function. In the case of `get_current_user`, there are no arguments (besides `self`, which needs no introduction), so these lines aren't necessary. Please remove them.

docs/current-user-ref.rst
```diff
@@ -4,3 +4,8 @@ CurrentUser
 
 .. autoclass:: canvasapi.current_user.CurrentUser
     :members:
+
+The CurrentUser class being a subclass of the User class, and thus inherits all of the methods of User,
+i.e the User class is the parent class from which methods are inherited to the Child(CurrentUser) class.
+
+See : https://canvas.instructure.com/doc/api/users.html for details
```
Thetwam, line 11: In the issue, by "provide a link to the `User` class reference page", I meant in the CanvasAPI documentation (as generated from `user-ref.rst`)

You should be able to do that with a link something like this:

```rst
`Documentation for the User class <user-ref.html>`_
```

Thetwam, line 11: Please move this section above the autoclass call

Thetwam (changes requested):
Thanks for the PR, @Xx-Ashutosh-xX! We appreciate the help 😁 

Just a few suggested tweaks below:

Xx-Ashutosh-xX (commented):
* I have removed the :params: and :type: from canvas.py
* Added the URL of User class liked you Suggested

Thetwam (approved):

Thetwam:
Great work, @Xx-Ashutosh-xX, thanks! 👍

PR #435, opened 2020-10-30
Issue/363 type hints

### LINKED ISSUE #363: types not exported from canvasapi prevents usage of type hints
# Describe the bug

Types other than `Canvas` are not exported from the canvasapi Python package. As a result, it is not possible to use [type hints](https://docs.python.org/3.7/library/typing.html) in our own code.

# To Reproduce

```python
import canvasapi

def get_course() -> canvasapi.Course:
    canvas = canvasapi.Canvas('...', '...')
    course = canvas.get_course(123456)
```

# Expected behavior

No error with a static type checker.

Instead, this error occurs using PyCharm's static type checker designed to work with Python type hints:

```
Cannot find reference 'Course' in '__init__.py' 
Inspection info: This inspection detects names that should resolve but don't. 
Due to dynamic dispatch and duck typing, this is possible in a limited but useful number of cases. 
Top-level and class-level items are supported better than instance items.
```

# Environment information

- Python version 3.7
- CanvasAPI version 0.15.0

bennettscience:
# Proposed Changes

Include type hints for the `<Canvas>` class. 

This is a first draft of including type annotations in the library. The contribution guidelines have also been updated to show how to add annotations to other modules following Python 3.6 formatting for backwards compatibility. I've tagged the original issue for reference, but this PR only includes annotations for `<Canvas>` at this point.

Fixes #363

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -28,8 +28,8 @@ and happy.
     -   [Type hinting](#type-hinting)
         -   [Type hint fomatting](#type-hint-formatting)
         -   [Type hints in Classes](#type-hints-in-classes)
-        -   [Python 3.6](#python-3.6)
-        -   [Python 3.7+](#python-3.7)
+        -   [Python 3.6](#python-3.6-type-hints)
+        -   [Python 3.7+](#python-3.7+-type-hints)
 
 ## How can I contribute?
 
@@ -451,16 +451,16 @@ def clear_course_nicknames(self):
     """
 ```
 
-### Type Hinting
+### Type Annotations
 
-Type hint support was added in Python 3.6 and allows maintainers to run static
-checks on code to identify and fix bugs caused by dynamic typing in scripts.
+Type annotations were added in Python 3.6 and allow maintainers to run static
+checks on code to identify and fix potential `TypeError` bugs before runtime.
 Type annotations should be added to maintain consistency across modules.
 
 #### Type hint formatting
 
-Type hints follow the general format below to define expected types for method
-parameters and returns.
+Type annotations follow the general format below to define expected types for
+method parameters and returns.
 
 ```python
 def my_function(param1: str, param2: str) -> str:
@@ -469,7 +469,7 @@ def my_function(param1: str, param2: str) -> str:
 
 #### Type Hints in Classes
 
-Classes instantiated with an **init**() function are expected to return `None`.
+Classes instantiated with an `__init__()` method are expected to return `None`.
 
 ```python
 class MyClass():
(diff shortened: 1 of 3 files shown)


bennettscience:
So, this failed because of the `mdl` check on spaces after list markers. The rule specifies 1, the contribution guidelines all use 3. Do you want to reformat for the standard rule or write our own for 3 spaces after a list marker?

.github/CONTRIBUTING.md
```diff
@@ -341,3 +450,124 @@ def clear_course_nicknames(self):
...
+
+        response = self.__requester.request(
+            "GET",
+            "appointment_groups/{}".format(appointment_group_id),
+            _kwargs=combine_kwargs(**kwargs),
+        )
+        return AppointmentGroup(self.__requester, response.json())
+```
+
+To get around this problem, the `typing` package has a `TYPE_CHECKING` module
```
jessemcbride, line 543: Is it possible to use [forward references](https://www.python.org/dev/peps/pep-0484/#forward-references) instead of conditional importing? I notice you [use them later](https://github.com/ucfopen/canvasapi/pull/435/files#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cR560), but I'm not sure if the conditional imports are a requirement for forward references to work or not.

canvasapi/canvas.py
```diff
@@ -230,7 +250,9 @@ def create_appointment_group(self, appointment_group, **kwargs):
 
         return AppointmentGroup(self.__requester, response.json())
 
-    def create_calendar_event(self, calendar_event, **kwargs):
+    def create_calendar_event(
+        self, calendar_event: dict, **kwargs: Union[str, dict, Optional[dict]]
```
jessemcbride, line 254: Matt and I have had a few conversations about this and we're not sure what the best approach is: how do we handle `kwargs`?

One of the nice things about taking in arbitrary keyword arguments is that canvasapi can maintain parity with Canvas updates as long as the endpoints themselves don't change. The parameters might change types or names or expect new values, but there's nothing we have to do downsteam.

Of course, the downside to that is that there's no useful information for developers in the function signature about what Canvas might accept. Typehints partially address this, but they reintroduce the API parity problem. 

Unless we can come up with a good tradeoff, I wonder if our `kwargs` typehint needs to be `Any` or something equally permissive.

canvasapi/canvas.py
```diff
@@ -498,7 +545,7 @@ def get_appointment_group(self, appointment_group, **kwargs):
         )
         return AppointmentGroup(self.__requester, response.json())
 
-    def get_appointment_groups(self, **kwargs):
+    def get_appointment_groups(self, **kwargs: Optional[dict]) -> PaginatedList:
```
jessemcbride, line 548: Is it possible to annotate with `PaginatedList[AppointmentGroup]`

jessemcbride (commented):
Hey @bennettscience, thank you for this patch. I'm sorry for the radio silence on it.

I've had an opportunity to experiment with mypy in some other projects and I think I finally have some useful feedback. Please take a look at the comments and let me know what you think when you get a chance.

.github/CONTRIBUTING.md
bennettscience: I'd tried that, but if I remember correctly, I had inconsistent results...mypy still threw errors. But, I could be remembering that wrong. I'll pull the last update back into this branch and see what happens.

canvasapi/canvas.py
bennettscience: Yeah, I did a lot of reading on how to type `**kwargs` and there isn't really a broad consensus. For me, I'm typically passing those in as a `dict`, but `Any` also makes sense. It's not considered bad practice because it really could be anything. Maybe a note in the docs somewhere about the `Any` type being limited to `kwargs` specifically?

canvasapi/canvas.py
bennettscience: Yes, you can do that directly or through a [Type Alias](https://docs.python.org/3/library/typing.html#type-aliases) to simplify complex returns.

PR #447, opened 2020-12-02
Troubleshooting Page

### LINKED ISSUE #414: Troubleshooting Guide
I think our docs (and maybe README?) would benefit greatly from a simple troubleshooting section. A majority of the questions we get about the library derive from the same handful user mistakes.

Items to include (but not limited to):
- Check URL formatting (provide details on exactly how URL should be formatted. Omit `/api/v1/`, trailing slash, etc.)
- Check access token (specify that it should be an access token and not a developer key, link to oauth2 process)

At the end of the troubleshooting section, we should also have a note for if the troubleshooting doesn't resolve the issue, prompting the user to create an Issue on GitHub or ask in the `#canvasapi` Slack channel (with links!).

Thetwam:
# Proposed Changes

* Add basic troubleshooting guide to documentation

Fixes #414.


jessemcbride (approved):
Looks great! I’m excited to keep expanding on this as we run into other common bugs. I think this will help folks out a lot.

PR #452, opened 2020-12-07
Fix return type for get_page and get_pages

### LINKED ISSUE #451: Incorrect documentation for Course.get_page() and Course.get_pages()
# Describe the bug

* The documentation for Course.get_page() says it returns a Course instead of a Page
* The documentation for Course.get_pages() says it returns a PaginatedList of Course instead of a PaginatedList of Page

Following this with a PR

hudcap:
# Proposed Changes

* change return type for Course.get_page() from Course to Page
* change return type for Course.get_pages() from PaginatedList of Course to PaginatedList of Page

Fixes #451

### DIFF
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -1794,7 +1794,7 @@ class Course(CanvasObject):
         :param url: The url for the page.
         :type url: str
         :returns: The specified page.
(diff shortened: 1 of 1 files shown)


Thetwam (approved):
Great catch, thanks! 👍

PR #457, opened 2021-01-05
Fix missing uri_str in Favorite (#456)

### LINKED ISSUE #456: Favorite object removal does not handle parameters (from response json data) correctly
# Describe the bug

_Favorite_ objects created using the `add_favorite_course` or `add_favorite_group` methods of the _CurrentUser_ class acquire a _context_type_ parameter that is capitalized (i.e. "Course" or "Group"). The `remove` method of the _Favorite_ class then makes use of this _context_type_ erroneously, testing for equality with _lowercase_ strings "course" or "group" (in the file canvasapi.favorite). As a result, the variable `uri_str` is not defined as it should be.

# To Reproduce

1. Instantiate a `canvasapi.canvas.Canvas` object using auth token: `c = Canvas(<base_url>, <token>)`
2. Acquire the CurrentUser object: `user = c.get_current_user()`
3. Use any valid "course_id" with the CurrentUser method `add_favorite_course` to acquire a Favorite object: `f = user.add_favorite_course(<course_id>)` (this will result in an object with `context_type` of "Course" (with a capital C)
4. Call the `remove` method of this Favorite object: `f.remove()` (results in an error because neither conditional was true, and so `uri_str` is undefined).

# Expected behavior

Course (or group) should be removed from favorites with no error.

# Environment information

- Python version: 3.8.1
- CanvasAPI version: 2.1.0

# Additional context

Canvas's HTTP API is presumably going to continue returning capitalized versions of "Course" and "Group" as their "context_type". Therefore, this issue could be fixed in in the file canvasapi.favorite (in the `remove` method) by:
- Comparing `self.context_type` with capitalized versions of "Course" and "Group"
- Adding the string `.lower()` method to `self.context_type` (to coerce the strings to lowercase for comparison).

Thank you! I have appreciated your package a lot!

deundrewilliams:
Compares the lowercased context_type to build the uri string

# Proposed Changes

* Call lower() on  self.context_type for comparison

Fixes #456


jessemcbride:
Hey Deundre,

This patch looks good so far, but let's add some regression tests to it to make sure this behavior works correctly. We have [two existing tests](https://github.com/ucfopen/canvasapi/blob/develop/tests/test_favorite.py#L39-L53) that are currently passing because they [receive a lowercased context type from our API fixture](https://github.com/ucfopen/canvasapi/blob/develop/tests/fixtures/favorite.json). We should add additional fixtures that return capitalized variants of `Group` and `Course` and test that a call to `favorite.remove()` returns successfully.

deundrewilliams:
Hello,

I added the new fixtures and tests that return the capitalized variants and made sure that the methods still passed.

jessemcbride (changes requested):
@deundrewilliams  Your code change is totally fine, but I think our existing favorites tests are a little wonky, which is influencing future tests. 

Right now, the setUp method [does some work](https://github.com/ucfopen/canvasapi/blob/271eccfeaf3bd537948d63b11098ab84510d9f52/tests/test_favorite.py#L26-L28) to get the current user and then set up a favorite course and a favorite group. This works for the existing tests because they’re only focused on [removing the favorite course and group](https://github.com/ucfopen/canvasapi/blob/develop/tests/test_favorite.py#L39-L53) and asserting that we consume the response to our DELETE request and return the same favorite object that we had before.

Unfortunately, the bug here is part of the interaction between the creation of our favorite objects and how we check to see which endpoint to call when we’re removing it. We need more control over the creation step under test, so we should probably pull it out of the `setUp` hook to provide it with its own fixture (the “return capitalized context codes instead of uncapitalized” one). If we do that, I think I’d like to modify our existing tests, too. The good news is that there aren’t many of them.

Does that seem reasonable? @Thetwam Any thoughts?

deundrewilliams:
I think decoupling the capitalized and uncapitalized codes is a good idea @jessemcbride. I'll start working on that now.

Thetwam:
It looks like the `flake8` and `black` checks are failing.

Running `black canvasapi tests` should do the trick.

tests/fixtures/current_user_capitalized_context_types.json
```diff
@@ -0,0 +1,152 @@
...
+	},
+	"get_by_id": {
+		"method": "GET",
+		"endpoint": "users/self",
+		"data": {
+			"id": 1,
+			"name": "John Doe"
+		},
+		"status_code": 200
+	},
```
Thetwam, line 50: I believe this endpoint is pretty much unchanged from the one that's already in `current_user.json`, so we can just reuse that one.

tests/fixtures/current_user_capitalized_context_types.json
```diff
@@ -0,0 +1,152 @@
+{
```
Thetwam, line 1: I'm not sure that a separate file for capitalization is necessary. Appending `_cap_context_type` or something like that to the new fixture names, but in `current_user.json` is probably fine.

@jessemcbride thoughts?

Thetwam (changes requested):
Code change is good. Tests need a little love, but it's a weird situation. Run `black` and we'll be all set.

tests/fixtures/current_user_capitalized_context_types.json
jessemcbride: Agreed.

Thetwam (approved):

jessemcbride (approved):

Issue #371, opened 2020-06-09
File Quota Information


Thetwam:
**What resource needs additional coverage?**
Files

**What endpoints need to be covered?**
https://canvas.instructure.com/doc/api/files.html#method.files.api_quota

- [ ] GET /api/v1/courses/:course_id/files/quota
- [ ] GET /api/v1/groups/:group_id/files/quota
- [ ] GET /api/v1/users/:user_id/files/quota

PR #458, opened 2021-01-11
Fixes issue (#371) - Adds endpoint for file quota information

deundrewilliams:
- This adds support for the file quota endpoint provided by the canvas API


canvasapi/user.py
```diff
@@ -572,6 +572,25 @@ def get_files(self, **kwargs):
             _kwargs=combine_kwargs(**kwargs),
         )
 
+    def get_file_quota(self, **kwargs):
+
+        """
+        Returns the total and used storage quota for the user.
+
+        :calls: `GET /api/v1/groups/:group_id/files/quota \
```
Thetwam, line 580: Should be user

```suggestion
        :calls: `GET /api/v1/users/:user_id/files/quota \
```

Thetwam (changes requested):
Looks great! Just need to run `black canvasapi tests` and fix a docstring and it's good to go.

Thetwam (changes requested):
We like to keep the main files organized by alphabetizing methods. `get_file_quota` should come before `get_files`. You can see where the errors are by running `python scripts/alphabetic.py`.

If you want to run all linters, tests, etc. at once locally, you can also run `scripts/run_tests.sh`.

_(**Note**: this is not a common - or even recommended - practice elsewhere, we're just doing it in CanvasAPI to keep massive files like `course.py` manageable)_

canvasapi/course.py
```diff
@@ -1312,6 +1312,25 @@ def get_file(self, file, **kwargs):
         )
         return File(self._requester, response.json())
 
+    def get_file_quota(self, **kwargs):
+
+        """
```
jessemcbride, line 1317: A tiny nitpick, but [docstrings shouldn't have a blank line between them and the function signature](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings). It looks like it won't interfere with Python's ability to parse it as a docstring, but it's probably worth updating each function to eliminate the blank line for consistency's sake.

Thetwam (approved):

PR #461, opened 2021-02-02
Fix typos in assignment.py

theunkn0wn1:
# Proposed Changes

Resolves two minor typos in `assignment.py` that result in invalid syntax.

Discovered when running doctests over the codebase for an unrelated branch.


Thetwam (approved):
Great catch, thanks! 👍

Thetwam (approved):

PR #463, opened 2021-02-02
issue/97 accept path like objects for file uploads

### LINKED ISSUE #97: Accept path-like objects for file uploads
Under `canvasapi` 0.7.0, `canvasapi.upload.Uploader.__init__` allows two ways to identify the file to be uploaded: it may be given as a string filename, or as an already-open file object. However, there are other ways to describe paths to files besides just strings: what the Python documentation calls “[path-like objects](https://docs.python.org/3/glossary.html#term-path-like-object).” For example:

- From Python 3.4 onward, the [`pathlib`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath) module provides a rich API for manipulating filesystem paths as path-like objects rather than as strings.

- From Python 3.6 onward, the [`os.DirEntry`](https://docs.python.org/3/library/os.html#os.DirEntry) objects returned by [`os.scandir`](https://docs.python.org/3/library/os.html#os.scandir) are path-like objects.

As far as I can tell, the `Uploader` class will not behave correctly if constructed with a path-like object as its `file` argument. This should be improved so that path-like objects are treated like openable filenames, rather than being assumed to be already-open files.

<hr/>

Python 3.6 added several library features to make it easier to write code that transparently handles both string paths as well as path-like objects:

- A new [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) abstract base class makes it possible to recognize path-like objects in a generic way, rather than checking for specific concrete classes.

- A new [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath) function returns the string representation of a path-like object, or simply returns its argument unchanged if that argument is already a string.

- Improvements to [`os.open`](https://docs.python.org/3/library/os.html#os.open) and other standard library functions allow them to accept either path-like objects or strings as arguments.

The Python 3.6 release notes include [more information on how to use these features](https://docs.python.org/3/whatsnew/3.6.html#pep-519-adding-a-file-system-path-protocol).

Unfortunately, those Python 3.6 improvements still leave a gap: Python 3.4 and 3.5 have the `pathlib` module, but do not have `os.PathLike`, `os.fspath`, the improvements to `os.open`, etc. Supporting path-like objects in these earlier Python releases may be inconvenient, or perhaps not worth doing at all.

theunkn0wn1:
# Proposed Changes
Add `canvas.upload.FileOrPathLike` type, which represents any `os.PathLike`, `str`, or `io` based file like and path like objects.
Add support in `canvas.upload.Upload`  to accept any `FileOrPathLike`, as per #97 /
Add type annotations to all consumers of `file` arguments which are forwarded into `canvas.upload.Upload`.

Added two tests that use `pathlib.Path` objects to assert correctness.

Fixes #97 .

### DIFF
--- a/canvasapi/upload.py
+++ b/canvasapi/upload.py
@@ -1,13 +1,14 @@
 import json
 import os
+import io
 
 from canvasapi.util import combine_kwargs
 
 from typing import Union
 
-PathLike = Union[os.PathLike, str]
+FileOrPathLike = Union[os.PathLike, str, io.IOBase, io.FileIO]
 """
-A path-like object. May be either a :class:`os.PathLike` or a `str`
+A path or file-like object. May be either a :class:`os.PathLike`, a `str`, or a file-like object
 """
 
 
@@ -16,7 +17,7 @@ class Uploader(object):
     Upload a file to Canvas.
     """
 
-    def __init__(self, requester, url, file: PathLike, **kwargs):
+    def __init__(self, requester, url, file: FileOrPathLike, **kwargs):
         """
         :param requester: The :class:`canvasapi.requester.Requester` to pass requests through.
         :type requester: :class:`canvasapi.requester.Requester`
--- a/canvasapi/upload.py
+++ b/canvasapi/upload.py
@@ -1,6 +1,7 @@
 import json
 import os
 import io
+from pathlib import Path
 
 from canvasapi.util import combine_kwargs
 
@@ -47,8 +48,11 @@ class Uploader(object):
             and the JSON response from the API.
         :rtype: tuple
         """
-        self.kwargs["name"] = os.path.basename(file.name)
-        self.kwargs["size"] = os.fstat(file.fileno()).st_size
+        if isinstance(file, Path):
+            self.kwargs.update({"name": file.name, "size": file.stat().st_size})
+        else:
+            self.kwargs["name"] = os.path.basename(file.name)
+            self.kwargs["size"] = os.fstat(file.fileno()).st_size
 
         response = self._requester.request(
             "POST", self.url, _kwargs=combine_kwargs(**self.kwargs)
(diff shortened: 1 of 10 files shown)


theunkn0wn1:
Ok. strangely this works in the test suite but not in my downstream (fails in an actual upload). It appears the tests I added were not sufficient. I recommend not merging this yet til I figure out how to properly test this & fix the bug.

theunkn0wn1:
image
works in downstream now.
It appears the tests were accurate, and i had a stale version of this project in the test environment leading to my last message :facepalm:

canvasapi/upload.py
```diff
@@ -39,8 +47,11 @@ def request_upload_token(self, file):
             and the JSON response from the API.
         :rtype: tuple
         """
-        self.kwargs["name"] = os.path.basename(file.name)
-        self.kwargs["size"] = os.fstat(file.fileno()).st_size
+        if isinstance(file, Path):
+            self.kwargs.update({"name": file.name, "size": file.stat().st_size})
```
jessemcbride, line 51: From what I can see, I think this code is unreachable with our current public API. We open the file before we call `get_upload_token`, so it's never a `Path`.

jessemcbride:
Our coverage drops here because of the `isinstance(file, Path)` case, but I think we can just remove the condition based on the current public API. @theunkn0wn1 Does that make sense, or are we missing something?

canvasapi/upload.py
theunkn0wn1: I believe that's correct, this change can likely be reverted.
IIRC I added this check while debugging something and didn't remove it after.

theunkn0wn1:
Sounds logical to me. pushed a commit to address that point.

Thetwam (approved):
Thanks for your work on this! LGTM 👍

PR #477, opened 2021-03-25
Deploy to PyPI with GitHub Actions

Thetwam:
# Proposed Changes

* Set up GitHub Action to deploy to PyPI on tag


jessemcbride (approved):
Let's see how it goes!

Issue #424, opened 2020-10-02
Time Zone Support for Smart DateTimes


Thetwam:
When writing documentation for our "Smart DateTimes" feature (#423), I discovered that it only supports strings without a UTC offset ("Z" at the end).

This happens because in `canvas_object.py`, we use this monstrosity of a regex string to determine if it is a date and pull out the appropriate pieces of the date:

```python
DATE_PATTERN = re.compile("[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z")
```

There is almost certainly a better way to do this, probably using built-in methods on the datetime module.

Rework this feature to work with all valid ISO 8601 datetime strings.

PR #488, opened 2021-04-15
Issue/424 time zone support for smart date times

kailukaitisBrendan:
# Proposed Changes

This pr address issue #424 by adding support for any string in the ISO 8601 format, including time offsets. This includes the use of the arrow library which helps to make parsing the ISO string easier.


jonespm:
I don't think datetime can any format dates. You'd have to use python-dateutil, arrow, pendulum or other similar packages. 

I'm not sure what's the most popular. I've generally just used dateutil, since it's what's [recommended in the docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat) for parsing.

Thetwam (approved):

(sorry it took so long)

PR #506, opened 2021-08-11
Issue #429: Course Analytics calls return a json, should return a PaginatedList

### LINKED ISSUE #429: Course Analytics summary data returns a json, should return a PaginatedList
# Describe the bug

The methods to return analytics data only return a json value, but Canvas returns this as a paginated list for some of these methods.

# To Reproduce

Call methods like `course.get_course_level_student_summary_data()` on a course with many users. This method returns json so it's not possible to get the next set of data, only the first.

# Expected behavior

Expected that these methods return a PaginatedList so you can loop to the next record.

# Environment information

- Python version (`python --version`) 3.8.5
- CanvasAPI version (`pip show canvasapi`) 2.0.0

# Additional context

Review the [other analytics](https://canvas.instructure.com/doc/api/analytics.html) (and possibly other) methods like
```
get_course_level_participation_data
get_course_level_assignment_data
get_user_in_a_course_level_assignment_data
get_user_in_a_course_level_messaging_data
```

to see if these also need a Paginatedlist.

craigdsthompson:
Fixes #429 by having the `get_course_level_student_summary_data()` method return a PaginatedList, instead of json.

I've also investigated the five other course-level analytics methods as suggested by @jonespm:
```
get_course_level_assignment_data()
get_course_level_participation_data()
get_user_in_a_course_level_assignment_data()
get_user_in_a_course_level_messaging_data()
get_user_in_a_course_level_participation_data()
```

and found that none of them return `Link` headers for paginated results, so I believe `get_course_level_student_summary_data()` is the only one that needs to be fixed.


jonespm (approved):
This looks good to me. I looked at the other calls in analytics and none of the others looked to be paginated.

Thetwam (approved):

PR #522, opened 2021-11-01
#440: added util.py:obj_or_id_or_sis_str,  use it in course.py:enroll…

### LINKED ISSUE #440: Allow SIS IDs
It is common for users to want to use SIS IDs in lieu of canvas IDs. [Canvas provides a way to use these alternate IDs](https://canvas.instructure.com/doc/api/file.object_ids.html) with the syntax `<ID_TYPE>:<ID_VALUE>`, e.g. `sis_course_id:A1234`.

In some cases, like [Canvas.get_user()](https://canvasapi.readthedocs.io/en/stable/canvas-ref.html#canvasapi.Canvas.get_user) and [Canvas.get_account()](https://canvasapi.readthedocs.io/en/stable/canvas-ref.html#canvasapi.Canvas.get_account), we added special (optional) arguments like `id_type` and `use_sis_id`, respectively. These allow users to indicate that the ID being passed is something other than a Canvas ID.

It turns out that there are many more places this functionality would be useful. In particular, we received a request to add it to [Course.enroll_user()](https://canvasapi.readthedocs.io/en/stable/course-ref.html#canvasapi.course.Course.enroll_user) for the `user` parameter. This is different from our previous implementations because those only modified the URL, not individual parameters.

Instead of adding optional arguments like `id_type`, `use_sis_id`, it's probably most straightforward to allow users to pass any properly-formatted string as an ID (e.g. `"sis_course_id:A1234"`). This is currently not possible with any ID that goes through the `obj_or_id` to check. We'll need to modify `obj_or_id` to allow SIS IDs. There are also "special IDs" that Canvas lists. We actually have some exceptions for `self`, but there are others.

This raises the question of what kind of checks to put into place regarding "valid" SIS IDs. Should we only allow specific ID types (from a list of all the Canvas-accepted ones), or allow any string that the user passes? Should we limit the types to only relevant ones (e.g. not allowing `sis_course_id` when the ID should be a user)?

Curious for others thoughts on the matter.

Some action items:

- [ ] Confirm manually (curl/Postman/etc.) that the [Enroll a user endpoint](https://canvas.instructure.com/doc/api/enrollments.html#method.enrollments_api.create) accepts SIS IDs for the `enrollment[user_id]` parameter
- [ ] Determine which SIS ID and special ID types we're going to support
- [ ] Update `canvasapi.util.obj_or_id` to accept SIS IDs (and possibly other special IDs)
- [ ] Determine how to handle existing implementations (such as `id_type` in `Canvas.get_user()`)

ndegroot:
…_user to support 'sis_login_id:' syntax

# Proposed Changes

* in util.py add new function 'obj_or_id_or_sis_str'
* in courses.py change function 'enroll_user' to use this function instead of 'obj_or_id'

Fixes #440 Allow SIS ids. An alternative could be to change obj_or_id to accept 'sis_login_id:xxxx'.
in section.py there also an enroll_user function, not changed.

### DIFF
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -37,6 +37,7 @@ from canvasapi.util import (
     normalize_bool,
     obj_or_id,
     obj_or_str,
+    obj_or_id_or_sis_str
 )
 
 
@@ -659,7 +660,7 @@ class Course(CanvasObject):
         from canvasapi.enrollment import Enrollment
         from canvasapi.user import User
 
-        kwargs["enrollment[user_id]"] = obj_or_id(user, "user", (User,))
+        kwargs["enrollment[user_id]"] = obj_or_id_or_sis_str(user, "user", (User,))
 
         if enrollment_type:
             warnings.warn(
--- a/canvasapi/util.py
+++ b/canvasapi/util.py
@@ -166,6 +166,50 @@ def obj_or_str(obj, attr, object_types):
         raise TypeError("Parameter {} must be of type {}.".format(obj, obj_type_list))
 
 
+def obj_or_id_or_sis_str(parameter, param_name, object_types):
+    """
+    Accepts either an int (or long or str representation of an integer) or
+    a 'sis_*_id:[some id]' format string or an object. If it is an int or
+    a format string, return it.
+    If it is an object and the object is of correct type, return the object's id. Otherwise,
+    throw an exception.
+
+    :param parameter: int, str, long, or object
+    :param param_name: str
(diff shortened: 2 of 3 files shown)


ndegroot:
I've just checked my code again, run the tests and coverage on Python 3.11 against current state. Still working ok. Please give me feedback why this is still pending. Happy to rework it!

ndegroot:
Added same changes to section.py in `def section()`. And added tests.

PR #531, opened 2022-03-09
Better custom gradebook column CRUD

Rojods:
Without the `{"content": column_data}`, we get a nasty `500` error everytime this algorithm tries to execute. I was surprised to hit it the first time I executed my script. The other change is to be able to set the column data for the first time. When there is already some data in the custom column, the `get_entries` routine returns and then we can modify then. But if there's nothing there in the first place, it is impossible with the current code base to add anything to start with. Therefore I propose to add in the `CustomGradebookColumn` a update the column data; this way it will be possible to create entries as well as modify them.

# Proposed Changes

* kwargs improvement
* new method to update or create `ColumnData` which works even when they don't exist.

# Fixes
* 500 errors when trying to use the python API for updating custom gradebook column entries.

For the record, I've modified the source code locally and tried using it successfully.

### DIFF
--- a/canvasapi/custom_gradebook_columns.py
+++ b/canvasapi/custom_gradebook_columns.py
@@ -73,6 +73,34 @@ class CustomGradebookColumn(CanvasObject):
         )
 
         return response.json().get("reorder")
+    
+    def update_custom_column_entry(self, user_id, column_data, **kwargs):
+        """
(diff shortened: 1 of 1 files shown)


gqmaguirejr:
I would like to support Rodolfo's posting regarding missing functionality as it should be possible to create new content rather than just update existing content.

PR #532, opened 2022-03-10
[Bug fix] Update `edit_front_page` to use `wiki_page` instead of `**kwargs`

z3ht:
# Proposed Changes

* Ensure peoples using a Course's `edit_front_page` provide a `wiki_page` dict instead of `**kwargs`

### DIFF
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -606,15 +606,23 @@ class Course(CanvasObject):
         )
         return ExternalFeed(self._requester, response.json())
 
-    def edit_front_page(self, **kwargs):
+    def edit_front_page(self, wiki_page, **kwargs):
         """
         Update the title or contents of the front page.
 
         :calls: `PUT /api/v1/courses/:course_id/front_page \
         <https://canvas.instructure.com/doc/api/pages.html#method.wiki_pages_api.update_front_page>`_
 
-        :rtype: :class:`canvasapi.course.Course`
+        :param wiki_page: The page.
+        :type wiki_page: dict
+        :returns: The created page.
+        :rtype: :class:`canvasapi.page.Page`
         """
+        if isinstance(wiki_page, dict) and "title" in wiki_page:
+            kwargs["wiki_page"] = wiki_page
+        else:
+            raise RequiredFieldMissing("Dictionary with key 'title' is required.")
+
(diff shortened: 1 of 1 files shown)


z3ht:
**Low priority**  
A simple bypass of this bug is to provide a `**kwargs` called `wiki_page` which contains all of the required keys

canvasapi/course.py
```diff
@@ -606,15 +606,23 @@ def delete_external_feed(self, feed, **kwargs):
...
-        :rtype: :class:`canvasapi.course.Course`
+        :param wiki_page: The page.
+        :type wiki_page: dict
+        :returns: The created page.
+        :rtype: :class:`canvasapi.page.Page`
         """
+        if isinstance(wiki_page, dict) and "title" in wiki_page:
+            kwargs["wiki_page"] = wiki_page
+        else:
+            raise RequiredFieldMissing("Dictionary with key 'title' is required.")
```
Thetwam, line 624: I see nothing in the [official documentation for editing the front page of a course](https://canvas.instructure.com/doc/api/pages.html#method.wiki_pages_api.update_front_page) about `"title"` being a required parameter. I tried using the old version of the function without the title and it was able to edit other parts of the page without me needing to include the title.

Since this endpoint seems to serve double-duty as both edit and initial creation of a front page, I suppose it's likely that `title` is required when creating a new front page from scratch. However, I don't want to require a user who is only editing the page to include a title if they don't absolutely need to.

Therefore, instead of including wiki_page (and specifically `title`) as a required parameter, I propose keeping the function as-is, and allowing the user to use the [standard keyword argument syntax](https://canvasapi.readthedocs.io/en/stable/keyword-args.html) to include whatever they need to:

```
course.edit_front_page(
    wiki_page={
        "title": "Title Goes Here",
        "body": "Content goes here",
    }
)
```

Thetwam (changes requested):
Hey @z3ht, thanks for your contribution!

It looks like this change causes one of the tests to fail, and causes a drop in test coverage, which would need to be resolved before this can be merged.

Additionally, I have some comments below about this change that might make the change ultimately unnecessary. Let me know what you think. Thanks!

Thetwam:
Closing as wontfix due to inactivity and previous comments.

PR #543, opened 2022-08-27
Pass through optional parameters in get_profile()

breed:
# Proposed Changes

* Pass through kwargs to the server. Most calls do this. A useful kwargs for get_profile is include=["links"]


breed:
i'm a little unclear how to write a test case for this. l looked at the get_profile test case but i'm not sure how to mock the optional parameters.

bennettscience:
The update looks okay. The Canvas docs don't show any keys to use in the request, but testing `include[]=links` in Postman returned a `links` key in the response. 

As far as writing tests, none of the other methods which accept `kwargs` pass in values, so as long as your tests are passing locally, it should be okay. Since it's your first contribution, the automated checks won't run until @Thetwam or @jessemcbride manually run them.

breed:
i can confirm that it works. i'm using it in this project https://github.com/breed/kattis2canvas to hook kattis to canvas.

Thetwam (approved):
Nice catch. Thanks for the fix! 👍

PR #559, opened 2022-10-24
Issue #459 missing modules

### LINKED ISSUE #459: Find Missing Kwargs script doesn't check all files
When reviewing #455, I discovered that `scripts/find_missing_kwargs.py` is now finding missing `**kwargs` in `enrollment.py`. There is no difference between that file on the issue branch and on `develop`, yet the script finds the error **only** in the issue branch.

This behavior occurs both in Travis and when I run the script manually on my local environment.

It looks like the problem starts early on in the script:

```python
for _, module in inspect.getmembers(canvasapi, inspect.ismodule):
```

If I replace `_` with `module_name` and print out the name immediately after, the list of modules is considerably shorter than expected. I suspect it's related to some shenanigans with imports and how we set up the `canvasapi` module.

Thetwam:
Addresses #459 by altering the import strategy for files, and adding a script to ensure `inspect` has the ability to see all the modules.

# Proposed Changes

* Add script `find_missing_modules` that compares the files in the `canvasapi` dir against the modules that `inspect` finds, then fails and warns which modules appear to be missing. (This affects the `alphabetic` and `find_missing_kwargs` scripts)
* Move a bunch of imports from function scope to top of file (note: cannot move all imports as-is due to circular imports)
* Add missing kwargs in newly-detected files
* Alphabetize functions to comply with `alphabetic` in newly-detected files
* Some formatting tweaks

Fixes #459


Thetwam:
Added missing `**kwargs`

Thetwam:
Functions now in alphabetical order.

PR #575, opened 2022-12-09
Updating documentation to include references to AssignmentOverride an…

lafent:
…d to display functionality of same.

# Proposed Changes

* This adds a reference to the autoclass for Assignments to pull in the AssignmentOverride details
* Also adds a line break to the `assignments.py` docstrings so they parse properly (same structure used in AssignmentGroup.

Fixes # .


Thetwam (approved):
Nice catch! Thanks for the help! 👍

PR #595, opened 2023-03-13
Issue #594 terminate user sessions

### LINKED ISSUE #594: Terminate user sessions
The user module needs a method to terminate all user sessions.

[Docs](https://canvas.instructure.com/doc/api/users.html#method.users.terminate_sessions)

lucas-salas:
Added endpoint coverage for [terminate all user sessions.](https://canvas.instructure.com/doc/api/users.html#method.users.terminate_sessions)

Fixes #594 

The docs for the endpoint doesn't specify any return type. I don't know a lick of ruby but it looks like in the [source code ](https://github.com/instructure/canvas-lms/blob/d77dbf14f8a55826a09607e07be8ac3dd4bae2d8/app/controllers/users_controller.rb#L2202) a success will just return `"ok"`?

Also not sure if I need to configure my `black` formatter since it change 18 files, or if that's normal.

### DIFF
--- a/canvasapi/communication_channel.py
+++ b/canvasapi/communication_channel.py
@@ -109,7 +109,6 @@ class CommunicationChannel(CanvasObject):
         :rtype: :class:`canvasapi.notification_preference.NotificationPreference`
         """
         if isinstance(notification_preferences, dict) and notification_preferences:
-
             for key, value in notification_preferences.items():
                 try:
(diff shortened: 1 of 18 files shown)


Thetwam (approved):

The `black` issue was just us running an old version. It was resolved elsewhere and your PR is clean now. Thanks again!

PR #612, opened 2023-04-21
New Quizzes

### LINKED ISSUE #608: New Quizzes
**What resource needs additional coverage?**

New Quizzes

- [x] [Get a new quiz](https://canvas.instructure.com/doc/api/new_quizzes.html#method.new_quizzes/quizzes_api.show)
- [x] [List new quizzes](https://canvas.instructure.com/doc/api/new_quizzes.html#method.new_quizzes/quizzes_api.index)
- [x] [Create a new quiz](https://canvas.instructure.com/doc/api/new_quizzes.html#method.new_quizzes/quizzes_api.create)
- [x] [Update a single quiz](https://canvas.instructure.com/doc/api/new_quizzes.html#method.new_quizzes/quizzes_api.update)
- [x] [Delete a new quiz](https://canvas.instructure.com/doc/api/new_quizzes.html#method.new_quizzes/quizzes_api.destroy)

**Note:** These endpoints follow a different base URL than most other endpoints (`/api/quiz/v1/` instead of `/api/v1/`)

Related Issue: #609

Thetwam:
# Proposed Changes

* Add coverage for New Quizzes endpoints

**Note:** These endpoints follow a different base URL than most other endpoints (`/api/quiz/v1/` instead of `/api/v1/`)

Fixes #608

### DIFF
--- a/canvasapi/course.py
+++ b/canvasapi/course.py
@@ -24,6 +24,7 @@ from canvasapi.grading_period import GradingPeriod
 from canvasapi.grading_standard import GradingStandard
 from canvasapi.license import License
 from canvasapi.module import Module
+from canvasapi.new_quiz import NewQuiz
 from canvasapi.outcome_import import OutcomeImport
 from canvasapi.page import Page
 from canvasapi.paginated_list import PaginatedList
@@ -451,6 +452,27 @@ class Course(CanvasObject):
         module_json.update({"course_id": self.id})
 
         return Module(self._requester, module_json)
+
+    def create_new_quiz(self, **kwargs):
+        """
+        Create a new quiz for the course.
+
+        :calls: `POST /api/quiz/v1/courses/:course_id/quizzes \
+        <https://canvas.instructure.com/doc/api/new_quizzes.html#method.new_quizzes/quizzes_api.create>`_
+
+        :returns: The newly-created New Quiz object
+        :rtype: :class:`canvasapi.new_quiz.NewQuiz`
+        """
+        endpoint = "courses/{}/quizzes".format(self.id)
+
+        response = self._requester.request(
+            "POST",
+            endpoint,
+            _url=self._requester.original_url + "/api/quiz/v1/" + endpoint,
+            _kwargs=combine_kwargs(**kwargs),
+        )
+
+        return NewQuiz(self._requester, response.json())
 
     def create_page(self, wiki_page, **kwargs):
         """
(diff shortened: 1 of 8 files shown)


canvasapi/course.py
```diff
@@ -452,6 +453,29 @@ def create_module(self, module, **kwargs):
...
+
+        :returns: The newly-created New Quiz object
+        :rtype: :class:`canvasapi.new_quiz.NewQuiz`
+        """
+        endpoint = "courses/{}/quizzes".format(self.id)
+
+        response = self._requester.request(
+            "POST",
+            endpoint,
+            _url=self._requester.original_url + "/api/quiz/v1/" + endpoint,
```
jonespm, line 471: I wonder if instead of defining the` _url `on all these calls we could modify `requester.py` to handle both of these patterns? Would eliminate a little redundancy. Would probably need like

```
self.base_url = base_url + "/api/v1"
self.new_quizzes_url = base_url + "/api/quiz/v1"
```

And then in `requester.request` would have to have some way of determining if it's a new quiz. I feel like there will be more methods so having this a a little cleaner now could be of benefit. 

I'm not sure the best way to handle that though. The easy solution might be to re-use url? Just an idea, not sure if it makes it significantly better.

```
        :param _url: Optional argument to specify the base URL to use for the request.
            If set to "new_quizzes", the `new_quizzes_url` instance variable will be used.
            If set to any other value (including None), the `base_url` instance variable will be used.
            If this is selected and an endpoint is provided, the endpoint will be ignored
            and only the `_url` argument will be used.
```
```
        if _url == "new_quizzes":
            base_url = self.new_quizzes_url
        else:
            base_url = self.base_url
```

bennettscience: That's a good point. The same could be done for the `graphql` endpoint in the requester.

jonespm: Right, I thought about that a little when implementing that method but that was *only* a single call. This is repeated a few times for each endpoint. But that could also be made to use this method.

jonespm: This could also be done on a separate issue just to get this in quicker, but I think it should be improved before future new_quizzes get added.

Thetwam: #619

Thetwam: I'm going to merge this in for now, but have spun this comment off into its own issue: #619. Will seek to complete that issue before doing a full release.

PR #622, opened 2023-05-17
Course Create Discussion Attach Files #621

### LINKED ISSUE #621: Attach files when creating a new discussion topic
# Describe the bug

Unable to upload attachment files in `Course.create_discussion_topic`

# To Reproduce

See discussion #620 (thanks @SebasMeza)

# Possible Solutions

This is a symptom of a larger problem with file handling. The library doesn't currently allow for files to be passed as arbitrary keyword arguments. That's a feature we should probably look into at some point (in a new issue), but for the purposes of this issue we can use an optional argument `assignment` and the `file_or_path` util function to achieve the same effect, like in `Course.import_outcome`. Since `assignment` isn't mandatory, the implementation will be slightly different.

Thetwam:
Fix issue where Course.create_discussion_topic doesn't accept attachment files

Fixes #621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 - Added support for pagination with metadata when headers are missing (Thanks, [@bennettscience](https://github.com/bennettscience))
 
+### Bugfixes
+
+- Fixed an issue where `Course.create_discussion_topic` wouldn't accept attachment files.
+
 ## [3.1.0] - 2023-04-21
 
 ### New Endpoint Coverage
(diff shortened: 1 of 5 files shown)


bennettscience (commented):
Do tests need to be updated to check cases where the file isn't valid?

Thetwam:
> Do tests need to be updated to check cases where the file isn't valid?

Good call, updated.

bennettscience (approved):
Looks good to me 👍

PR #630, opened 2023-06-15
#629 Implement default request timeouts

### LINKED ISSUE #629: Request timeouts
Having a timeout functionality should be a living standard for API wrappers. I propose for there to be an interface that allows for setting request timeouts without complicated subclassing.

ArhanChaudhary:
# Proposed Changes

Adds a default API request timeout. I originally wanted to have a timeout keyword argument for every Requester.request call, but there were far too many instances of self._requester.request in the codebase, so I've instead insisted on a global default timeout.

I was unsure how to write the unit test for the mocked request wait and raise a ReadTimeout, so I'll just have to leave that up to a reviewer.

Fixes #629 .

### DIFF
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -72,7 +72,9 @@ class Canvas(object):
         access_token = access_token.strip()
         base_url = get_institution_url(base_url)
 
-        self.__requester = Requester(base_url, access_token, default_timeout=default_timeout)
+        self.__requester = Requester(
+            base_url, access_token, default_timeout=default_timeout
+        )
 
     def clear_course_nicknames(self, **kwargs):
         """
--- a/canvasapi/canvas.py
+++ b/canvasapi/canvas.py
@@ -33,12 +33,14 @@ class Canvas(object):
     The main class to be instantiated to provide access to Canvas's API.
     """
 
-    def __init__(self, base_url, access_token):
+    def __init__(self, base_url, access_token, default_timeout=None):
         """
         :param base_url: The base URL of the Canvas instance's API.
         :type base_url: str
         :param access_token: The API key to authenticate requests with.
         :type access_token: str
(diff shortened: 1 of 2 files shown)


bennettscience:
A couple of preliminary thoughts:

1. I like the idea of having a timeout, but the nuance of timeouts in requests may trip people up. [According to the docs](https://requests.readthedocs.io/en/stable/user/advanced/#timeouts), providing an `int` will apply to the `read` and `write` timeouts. Would a `tuple` make more sense so the user explicitly has to set both values?
2. Related, [this is not an upper limit on download time](https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts), just time to first bytes received. Does that matter?
3. Do we need to subclass `CanvasException` to handle the `[Timeout](https://requests.readthedocs.io/en/latest/api/#requests.Timeout)` emitted by `requests`? 
4. A unit test with a sleep function greater than the timeout set on `Canvas` would be fine to make sure the error is captured correctly.

ArhanChaudhary:
Hello @bennettscience, thanks for your thoughts. I'll commit and address your insights once I get feedback on my response:

1. Sure, as it's consistent with the requests module.
2. No, for the sake of consistency. Considering how widely used the requests module is for example with other API wrappers, having a timeout functionality that has consistent behavior is easier to understand. On second thought, maybe it's better to rename default_timeout to just timeout to emphasize that its interface mirrors the requests module's.
3. Definitely. I didn't think about that initially. Though I would prefer if it instead subclassed ReadTimeout for safety, as that's what's explicitly thrown when the request timeout exceeds. 
4. Agreed. Would it work if I add a `exc=None` kwarg to `register_uris` and pass that into `requests_mocker.register_uri` so `ReadTimeout`s can be caught in a `with self.assertRaises(ReadTimeout):` block?

PR #652, opened 2024-05-10
Issue/619 Improve new quizzes & graphql requester

### LINKED ISSUE #619: Rework url structure to accommodate graphql and new quizzes endpoints
> I wonder if instead of defining the` _url `on all these calls we could modify `requester.py` to handle both of these patterns? Would eliminate a little redundancy. Would probably need like
>
> ```
>self.base_url = base_url + "/api/v1"
>self.new_quizzes_url = base_url + "/api/quiz/v1"
>```
>
>And then in `requester.request` would have to have some way of determining if it's a new quiz. >I feel like there will be more methods so having this a a little cleaner now could be of >benefit. 
>
>I'm not sure the best way to handle that though. The easy solution might be to re-use url? >Just an idea, not sure if it makes it significantly better.
>
>```
>        :param _url: Optional argument to specify the base URL to use for the request.
>            If set to "new_quizzes", the `new_quizzes_url` instance variable will be used.
>            If set to any other value (including None), the `base_url` instance variable will >be used.
>            If this is selected and an endpoint is provided, the endpoint will be ignored
>            and only the `_url` argument will be used.
>```
>```
>        if _url == "new_quizzes":
>            base_url = self.new_quizzes_url
>        else:
>            base_url = self.base_url
>```

_Originally posted by @jonespm in https://github.com/ucfopen/canvasapi/pull/612#discussion_r1185321160_

bennettscience:
# Proposed Changes
Add a `new_quizzes` and `graphql` keyword to the `Requester` instance to handle special API endpoints.

Fixes #619.


bennettscience:
Closing because I need to clean up my commit history.

PR #653, opened 2024-05-11
Use `new_quizzes` and `graphql` keywords

### LINKED ISSUE #619: Rework url structure to accommodate graphql and new quizzes endpoints

bennettscience:
Based on #612 and as an issue in #619, this adds the `new_quizzes` and `graphql` keyword endpoints in the requester module. Methods which specify these locations will have the correct URL supplied by the library in the `_url` parameter for the requester.

Calls from the `canvas`, `course`, and `new_quizzes` modules have been updated to use they keywords rather than a formatted URL string.

All tests passing.

Fixes #619.

### DIFF
--- a/canvasapi/requester.py
+++ b/canvasapi/requester.py
@@ -36,7 +36,7 @@ class Requester(object):
         self.original_url = base_url
         self.base_url = base_url + "/api/v1/"
         self.new_quizzes_url = base_url + "/api/quiz/v1/"
-e       self.graphql = base_url + "/api/graphql"
+        self.graphql = base_url + "/api/graphql"
         self.access_token = access_token
         self._session = requests.Session()
         self._cache = []
@@ -172,9 +172,9 @@ e       self.graphql = base_url + "/api/graphql"
         if not _url:
             full_url = "{}{}".format(self.base_url, endpoint)
         elif _url == "new_quizzes":
-            full_url = "{}{}"".format(self.new_quizzes_url, endpoint)
+            full_url = "{}{}".format(self.new_quizzes_url, endpoint)
         elif _url == "graphql":
-            full_url_url = self.graphql
+            full_url = self.graphql
         else:
             full_url = _url
 
--- a/canvasapi/requester.py
+++ b/canvasapi/requester.py
@@ -35,6 +35,8 @@ class Requester(object):
         # Preserve the original base url and add "/api/v1" to it
         self.original_url = base_url
         self.base_url = base_url + "/api/v1/"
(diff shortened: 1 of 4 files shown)


canvasapi/requester.py
```diff
@@ -159,7 +167,16 @@ def request(
...
+        # Check for specific URL endpoints available from Canvas. If not
+        # specified, pass the given URL and move on.
+        if _url == "new_quizzes":
+            request_url = "{}{}".format(self.new_quizzes_url, endpoint)
+        elif _url == "graphql":
+            request_url = self.graphql
+        else:
+            request_url = _url
+
+        full_url = request_url if _url else "{}{}".format(self.base_url, endpoint)
```
Thetwam, line 179: Since we're doing multiple checks now, the inline if/else is a bit harder to read. How about something like this?

```suggestion
        if not _url:
            full_url = "{}{}".format(self.base_url, endpoint)
        elif _url == "new_quizzes":
            full_url = "{}{}".format(self.new_quizzes_url, endpoint)
        elif _url == "graphql":
            full_url = self.graphql
        else:
            full_url = _url
```

Thetwam (changes requested):

canvasapi/requester.py
bennettscience: Yep...makes more sense to check if the argument is even included first.

Thetwam (approved):
Looks good. Thanks! 👍

PR #654, opened 2024-05-11
Improve `PaginatedList` docs

### LINKED ISSUE #651: Improve `PaginatedList` docs
The `PaginatedList` page in the docs doesn't have anything here. There are several parameters that are helpful (like `_root` when fetching nested response data) that don't have any details.

bennettscience:
Include details for each parameter used in `PaginatedList`.

Fixes #651.

### DIFF
--- a/canvasapi/paginated_list.py
+++ b/canvasapi/paginated_list.py
@@ -5,7 +5,6 @@ class PaginatedList(object):
     """
     Abstracts `pagination of Canvas API \
     <https://canvas.instructure.com/doc/api/file.pagination.html>`_.
-    Returns a complete list of all results of a specified type.
     """
 
     def __getitem__(self, index):
--- a/canvasapi/paginated_list.py
+++ b/canvasapi/paginated_list.py
@@ -5,6 +5,7 @@ class PaginatedList(object):
     """
     Abstracts `pagination of Canvas API \
     <https://canvas.instructure.com/doc/api/file.pagination.html>`_.
+    Returns a complete list of all results of a specified type.
     """
 
     def __getitem__(self, index):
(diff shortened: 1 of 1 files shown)


bennettscience:
I'm not sure why the `flake8` check is failing - it's passing locally.

jonespm:
Looks like some error with the Codecov action not with this PR. I'm don't think I have the permission to look into this any more. Maybe it'll pass when run tomorrow or Monday.

Thetwam:
Think I fixed the Codecov thing in #655 and by updating the upload token. (Not sure if it was one, the other, or both, but hey).

Get that in and this should hopefully work

canvasapi/paginated_list.py
```diff
@@ -5,6 +5,7 @@ class PaginatedList(object):
     """
     Abstracts `pagination of Canvas API \
     <https://canvas.instructure.com/doc/api/file.pagination.html>`_.
+    Returns a complete list of all results of a specified type.
```
Thetwam, line 8: This is a bit misleading since PaginatedList doesn't actually return a list, let alone a complete one. It kinda grabs what it needs while it goes.

Maybe mention that it's an iterator, or something along those lines?

Thetwam (changes requested):
Thanks for adding docs! The docstring on `__init__` is super useful.

I'm concerned the addition to the class docstring may be misleading.

canvasapi/paginated_list.py
bennettscience: I went back and forth on that last night. I ended up just taking it out because just having `__init__` documented will help.

Thetwam (approved):
Thanks! 👍

Issue #305, opened 2019-10-03
Negative index on PaginatedList causes unexpected results


UniversalSuperBox:
# Describe the bug

It is expected that you cannot negative-index the PaginatedList due to its inability to know its length until it is full. However, it is still possible to use the negative index if at least one page of results has been loaded. When this happens, the list returns the last value it has loaded (which is not necessarily the last value it can load).

# To Reproduce

```python
users = canvas.get_course(1).get_users()
users[1]
users[-1]
```

# Expected behavior

I am not allowed to use negative indexing (wishlist, I *can* use negative indexing :D)

# Environment information (please complete the following information)

- Python 3.7.0
- CanvasAPI 0.14.0

# Additional context

I'll try my hand at fixing this.

PR #237, opened 2019-04-08
#114 Implement __len__() for paginated_list

onomou:

Issue #175, opened 2018-05-04
Add coverage for retrieving an activity stream


jessemcbride:
Right now we wrap [get_activity_stream_summary()](https://canvas.instructure.com/doc/api/users.html#method.users.activity_stream_summary) but not [get_activity_stream()](https://canvas.instructure.com/doc/api/users.html#method.users.activity_stream).

We should include that.

Thetwam:
This endpoint isn't covered due to a limitation of our current implementation of `PaginatedList`. `PaginatedList` expects objects of a single type, but this endpoint returns objects of mixed types. To cover this endpoint, we'll need to improve `PaginatedList`.

Issue #114, opened 2017-11-27
PaginatedList redux


jessemcbride:
Modify `PaginatedList` to provide support for special methods on some `CanvasObject` types:

* `len`
* negative indexing

Thetwam:
See discussion on https://github.com/ucfopen/canvasapi/pull/237#pullrequestreview-227354435 for limitations on implementation

jessemcbride:
See https://gist.github.com/jessemcbride/133312e5af9e6eb1de7d5b77dc9b7ba9 for implementation scaffolding.

jonespm:
Was this still being worked on? I think just knowing the number of pages and items in a list, which is often available from the last link and available via a calculation of per_page would be really useful for api calls that have tons of pages. The scaffolding example is kind of how I was thinking this would go. I'm less concerned about the negative index than just being able to estimate how long processing will take. Thanks for adding this!

PR #656, opened 2024-05-15
Issue/539 typed paginated list

bennettscience:
Original PR (#539) from garth74:

> This pull request makes two changes:
>
> Adds type hinting to PaginatedList so that elements used in for-loops, list comprehensions, etc. are typed ([more info](https://mypy.readthedocs.io/en/stable/generics.html)).
> Adds negative indexing to PaginatedList.
>
> Related
> Supports efforts of https://github.com/ucfopen/canvasapi/pull/435 and fixes https://github.com/ucfopen/canvasapi/issues/305 without preventing negative indexing all together (i.e., https://github.com/ucfopen/canvasapi/pull/306).

Updated to include upstream changes in `develop` since the original was opened in 2022.

### DIFF
--- a/canvasapi/paginated_list.py
+++ b/canvasapi/paginated_list.py
@@ -4,6 +4,8 @@ import re
 import typing as t
 from itertools import islice
 
+from requests.models import Response
+
 if t.TYPE_CHECKING:
     from canvasapi.requester import Requester  # pragma: no cover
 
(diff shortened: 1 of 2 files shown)


bennettscience:
I think there are two things worth discussing:

1. The typing improvements will help. I think `PaginatedList` is one of the harder bits of the library to get right if you've never used it before. It might help users in more complex cases avoid gotchas.
2. My concern is about negative indexing large lists. It still has to get all of the results before you can access the results. Should we (I?) put in some more effort to address the other related issues (see #114, #237, and #175.)

PR #672, opened 2024-12-02
Issue/669 GraphQL Updates to PaginatedList

### LINKED ISSUE #669: GraphQL needs to be put in PaginatedList
# Describe the bug

An upcoming (or recent depending on when you read this) [change coming to Canvas](https://community.canvaslms.com/t5/Canvas-Change-Log/2024-GraphQL-API-Change-Log/ta-p/617064) is going to impose per page limits to GraphQL requests, previously unlimited. GraphQL call isn't paginated in this code so it likely won't work.

# To Reproduce

1. Need to setup CanvasAPI to run a GraphQL call
2. Will fill in more details later once it's tested but any call that returns more than 10 results (like all courses for a term) would likely only return the first page. 

# Expected behavior

Return a paginated list so the user can page through it.

# Environment information

- Python version 3.12
- CanvasAPI version 3.3.0

jsmnhou:
# Proposed Changes
* Updates PaginatedList object (in paginated_list.py) to set default page size (per_page parameter) to 20 (previously 100) to adhere to [10/2024 GraphQL API changes](https://community.canvaslms.com/t5/Canvas-Change-Log/2024-GraphQL-API-Change-Log/ta-p/617064)
* Add associated tests/fixtures (see `test_paginated_list.py`, `fixtures/paginated_list.json`)

Fixes #669

# Effects
* With these changes, all tests are passing and coverage report remains at 100%.

### DIFF
--- a/canvasapi/paginated_list.py
+++ b/canvasapi/paginated_list.py
@@ -52,7 +52,7 @@ class PaginatedList(object):
         self._content_class = content_class
         self._first_url = first_url
         self._first_params = kwargs or {}
-        self._first_params["per_page"] = kwargs.get("per_page", 100)
+        self._first_params["per_page"] = kwargs.get("per_page", 20)
         self._next_url = first_url
         self._next_params = self._first_params
         self._extra_attribs = extra_attribs or {}
--- a/tests/fixtures/account.json
+++ b/tests/fixtures/account.json
@@ -1885,7 +1885,7 @@
 	},
 	"get_authentication_events": {
 		"method": "GET",
-		"endpoint": "audit/authentication/accounts/1?per_page=100",
+		"endpoint": "audit/authentication/accounts/1?per_page=20",
 		"data": [
 			{
 				"created_at": "2012-07-19T15:00:00-06:00",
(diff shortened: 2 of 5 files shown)


jonespm:
Hi @jsmnhou thanks for taking a look at this!

So the paginated list limit change (which was also currently delayed) was only targeted at the _GraphQL_ API, not all API's. So unfortunately the code you contributed wouldn't fix this problem as we don't want the default changed for the other API's.

GraphQL is a special separate API and actually has it's own way of doing pagination using the [Relay Connection Spec](https://relay.dev/graphql/connections.htm) as [mentioned on this documentation](https://canvas.instructure.com/doc/api/file.graphql.html). It adds new `first`/`after `arguments and It doesn't use response headers like the other API's. I was thinking that the code here could be a method like `fetch_all_pages `and programmatically inject these and make the call. I can add some more examples to the original issue.

So it would need completely new custom code somewhere to handle this, maybe something like a paginated_graphql_list.py or maybe just in util.

The `graphql` call is in canvas.py https://github.com/ucfopen/canvasapi/blob/5e4f2600247116a46701ef26ce046a213b3df519/canvasapi/canvas.py#L1261 sends the request "as-is". 

I haven't looked at it at all, just filed this one as an issue.  But unfortunately this fix wouldn't solve the problem.

jsmnhou:
@jonespm That makes sense, thank you so much for the reply! It definitely seems like I was on the wrong track. I'll go ahead and close this PR for now and open a new one if I get the chance to take a second look!

PR #680, opened 2025-01-03
find_missing_modules should import rather than append to prioritize local

### LINKED ISSUE #675: Find Missing Modules should import rather than append to prioritize local module
# Describe the bug

When running the test `find_missing_modules.py`, it will fail if a new module is added locally while an older version of CanvasAPI is installed in the path that doesn't have this. This is because the local path has the lower priority. I don't think this will be a problem with Github action

# To Reproduce

* Install a version of CanvasAPI globally or in the virtual environment
* Add a new module, also import it in another module as required. 
* Attempt to run the `scripts/find_missing_modules.py`

Script should pass if done correctly. It fails. If the global canvasapi is removed this test does pass.

# Expected behavior

The local `canvasapi` should always be preferred by this test

`sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))`

jonespm:
# Proposed Changes

* Just change the script action to import rather than append which will put the local path first

Fixes #675

### DIFF
--- a/scripts/find_missing_modules.py
+++ b/scripts/find_missing_modules.py
@@ -2,7 +2,7 @@ import inspect
 import os
 import sys
 
-sys.path.append(os.path.join(sys.path[0], ".."))
+sys.path.insert(0,(os.path.join(sys.path[0], "..")))
 
 import canvasapi  # noqa
 


bennettscience (approved):
I haven't run into this bug before, but this looks good to me.

PR #686, opened 2025-03-21
fix: don't override `per_page` in `PaginatedList`

### LINKED ISSUE #685: PaginatedList doesn't process per_page and other kwargs
# Describe the bug
PaginatedList hasn't been updated for the `combine_kwargs` method. It expects `**kwargs` being formatted as 
`PaginatedList(..., **kwargs)` instead of `PaginatedList(...,_kwargs=combine_kwargs(**kwargs))` or `PaginatedList(..., kwargs=combine_kwargs(**kwargs))`.

Some functions use `_kwargs=..` and others use `kwargs=..`,  but the requester method uses `_kwargs` as a input so I am guessing `_kwargs` should be used for all.

currently this is done
```python
        self._first_params = kwargs or {}
        self._first_params["per_page"] = kwargs.get("per_page", 100)
```
where 100 is the default instead of the 10 mentioned in the documentation.

# To Reproduce
```python
import time
CANVAS_TOKEN = ...; API_URL = ...;
canvas = Canvas(CANVAS_TOKEN, API_URL)
start = time.perf_counter()
list(canvas.get_courses(per_page=1))
print(time.perf_counter() - start)
start2 = time.perf_counter()
list(canvas.get_courses(per_page=30))
print(time.perf_counter() - start2)
```

# Expected behavior

one would take longer then the other, but they take the same since both have per_page set to 100.
`kwargs` will be `{'_kwargs': [('per_page', 1)]}` or `{'_kwargs': [('per_page', 30)]}`, but since `per_page` is not in it, the default is used and 
`self._first_params = {'_kwargs': [('per_page', 1)], 'per_page': 100}`

# Environment information
it is still in the latest version of the github repo, 
- CanvasAPI version 3.3.0

# Additional context

superle3:
# Proposed Changes

* Don't override `per_page` parameter
* update readme to state 100 is used instead of the 10 default when it is not specified, but we specify it already by default
* add _kwargs as arg to `PaginatedList` and change every `kwargs=combine_kwargs(**kwargs)` to  `_kwargs=combine_kwargs(**kwargs)`, since I couldn't find a reason why some were different in the documentation and commits.

Fixes #685 .

TODO: a test maybe?


superle3:
fixtures were wrong, updated them and the tests accordingly.

PR #705, opened 2025-12-17
RateLimitExceeded

### LINKED ISSUE #704: Update RateLimitExceeded Exception
Canvas has (finally) made rate limit exceeding errors throw a 429 instead of a 403. Update accordingly.

Thetwam:
# Proposed Changes

* Update rate limit exception to happen on 429

Fixes #704

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### General
+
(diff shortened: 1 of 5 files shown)


jonespm (approved):
I don't think we have anything that hits the rate limit currently using this, but it's nice that it's distinct and PR looks good.

PR #707, opened 2025-12-23
Add User-Agent

### LINKED ISSUE #706: User-Agent needed from January 2026
https://community.instructure.com/en/discussion/658205/enforcing-user-agent-header-for-canvas-api-requests

Starting in January 2026, Canvas requires the User-Agent header on API requests. Looking quickly, I couldn't find any indication of this being set with canvasapi. But I guess we should set a default and add a way to set a custom one easily.

Thetwam:
# Proposed Changes

* Add a default User-Agent header

Fixes #706


dgwn (approved):
Tested locally in the Python shell against our local Canvas instance, LGTM 👍

PR #713, opened 2026-03-12
Fix QuizGroup's course_id

### LINKED ISSUE #712: Quiz.create_question_group returns QuizGroup with wrong course_id
# Describe the bug

The QuizGroup returned from create_question_group has a course_id of the quiz_id.

# To Reproduce

1. Get a quiz object (call it `quiz`)
2. Create a question group with `create_question_group`:
`new_group = quiz.create_question_group(quiz_groups=[{'name':'New Group'}])`
3. Access `course_id` and compare to quiz course_idwill evaluate to True only if course_id == quiz_id, should always evaluate to True
`new_group.course_id == quiz.course_id`

# Expected behavior

Expected that course_id should be set the course_id, and not the quiz_id.

# Environment information

- Python version 3.14
- CanvasAPI version 3.4.0

# Additional context

Current line 115 of quiz.py is:
`response_json["quiz_groups"][0].update({"course_id": self.id})`
but should be
`response_json["quiz_groups"][0].update({"course_id": self.course_id})`

dgwn:
# Proposed Changes

* `QuizGroup` fixed to have `course_id` of the quiz's `course_id`, instead of the `quiz_id`

Fixes #712.


Thetwam (approved):
Thanks!

